PR #9, opened 2016-12-05
Fix event list initial loading to determine the need based on retaine…

### LINKED ISSUE #5: Blank screen in list view.
Every now and then I get a blank list when I resume the app. Closing the app and re-opening reloads the list.

lemonboston:
…d objects and not on savedInstanceState. Some other minor refactor. Fixes #5.

@dmfs Please review.


lemonboston:
The _Don't keep Activities_ debug option can be used to test this. 
Before the fix, the screen was blank, after the fix the screen reloads the events.

Issue #4, opened 2016-12-02
Remove dependency on 'com.github.SmoothSync:smoothsetup'


lemonboston:
Just a couple of classes are used from that unrelated project, we should copy them here for now (maybe creating separate library later), and remove the dependency.

Note: `ThrowingAsyncTask` usage can also be replaced with `SafeAsycnTask` that is already in this project.

dmfs:
Done.

PR #10, opened 2016-12-06
Remove dependency on SmoothSetup by copying FutureLocalServiceConnect…

dmfs:
…ion to this project and using SafeAsyncTask rather than ThrowingAsyncTask

implements #4

@lemonboston please review

### DIFF
--- a/eventdiscovery-sdk/build.gradle
+++ b/eventdiscovery-sdk/build.gradle
@@ -38,7 +38,6 @@ dependencies {
     compile 'org.dmfs:http-client-types:0.10.2'
     compile 'org.dmfs:http-executor-decorators:0.10.2'
     compile 'org.dmfs:httpurlconnection-executor:0.10.2'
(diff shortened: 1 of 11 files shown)


lemonboston:
@dmfs Look good to me.

Just some notes on the AsyntTasks: in this `SafeAsyncTask` the idea with separating the task and execution parameters is also to be in align with the declarative approach where constructor parameters are what need to define that intent, and method params are for the execution only. So in current `ActionLoaderTask` `FutureServiceConnection` would be execution param and `Event[]` would be task param.
But I don't like the vararg plus generic part, so there is probably room to improve this, maybe creating multiple types of these SafeAsyncTasks for convenience, with different number of params, or allow to be interested in the task params in the callback or not, or simply not use task param as in current `ActionLoaderTask`.

dmfs:
thx. We'll consider that on the next refactoring. I'll keep it like that for now.

PR #19, opened 2016-12-07
PR for 18 - loading indicator items

### LINKED ISSUE #18: Event list: loading indicator top/bottom item instead of the overlay
@lemonboston commented on [Thu Dec 01 2016](https://github.com/schedjoules/android-event-discovery-sdk-int/issues/51)

Instead of the loading indicator displayed in the middle we could add an item to the bottom(/top) of the list while the loading is ongoing. It's probably nicer and also because loading next page starts automatically when user scrolls within a distance to bottom, he wouldn't even see it most of the case.

---

@lemonboston commented on [Thu Dec 01 2016](https://github.com/schedjoules/android-event-discovery-sdk-int/issues/51#issuecomment-264137161)

Technical note:
The removing of special items (error, no more events, loading indicator) to top/bottom needs to be revised to not use only the first/last position to make it robust.

### LINKED ISSUE #51: In order to comply with 3rd party open source licenses, as a partner I need a list of open source libs used by the SDK.
We should add the list of dependencies and their licenses to the wiki.

lemonboston:
Resolves #18 .

@dmfs Please review.


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/EventListActivity.java
```diff
@@ -121,7 +120,7 @@ private void setupListAdapter()
         AdapterNotifier adapterNotifier = new FlexibleAdapterNotifier(adapter);
 
         // Using GeneralMultiTypeAdapter (no sticky headers):
```
dmfs: please remove the commented sections

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/items/LoadingIndicatorItem.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public final class LoadingIndicatorItem extends AbstractFlexible<View> implements ListItem<View>
+{
+    public static final LoadingIndicatorItem TOP = new LoadingIndicatorItem();
+    public static final LoadingIndicatorItem BOTTOM = new LoadingIndicatorItem();
+
+
+    public static LoadingIndicatorItem get(ScrollDirection direction)
+    {
+        return direction == ScrollDirection.BOTTOM ? BOTTOM : TOP;
```
dmfs: Consider replacing this static method by something like 
```java
public enum ScrollDirection
{
    TOP(new LoadingIndicatorItem()),
    BOTTOM(new LoadingIndicatorItem());
    
   ListItem<View> mListItem;
    public ScrollDirection(ListItem<View> i)
   {
      mListItem = i;
    }

   public ListItem<View> listItem()
   {
      return mListItem;
   }
}
```

That way you can just say 
```java
   ScrollDirection.TOP.listItem();
```
instead of 
```java
    LoadingIndicatorItem.get(ScrollDirection.TOP);
```

This might need a few changes (like a proper factory interface) to work well but you get the idea.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/itemsprovider/EventListItemsProviderImpl.java
```diff
@@ -81,18 +80,15 @@
...
         mApiService = apiService;
         mBackgroundMessage = backgroundMessage;
         mItems = items;
-        mBackgroundMessage.setOnClickListener(this);
         mLoadingIndicatorOverlay = loadingIndicatorOverlay;
+        mBackgroundMessage.setOnClickListener(this);
         mExecutorService = Executors.newSingleThreadExecutor();
         mDownloadTaskClient = new DownloadTaskClient();
-        mLastResultPage = new HashMap<>(2);
         mErrorTaskParam = new HashMap<>(2);
```
dmfs: Creating  HashMap with size 2 can only hold 1 element without having to rehash. Is that really what you want?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/itemsprovider/EventPageQueries.java
```diff
@@ -18,60 +18,75 @@
...
     {
-        ResultPage<E> resultPage = mResultPages.get(mDirection);
+        ResultPage<Envelope<Event>> resultPage = mLastResultPages.get(direction);
         if (resultPage == null)
         {
             return false;
         }
 
-        switch (mDirection)
+        switch (direction)
```
dmfs: IMO doing a Switch with enum is bad style in most cases.
Why not let the enum handle this like so?
```
   return direction.isAvailable(mLastResultPages.get(direction));
```
No Switch, no (stale) Exception.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/itemsprovider/EventPageQueries.java
```diff
@@ -18,60 +18,75 @@
...
+    public ApiQuery<ResultPage<Envelope<Event>>> comingPageQuery(ScrollDirection direction) throws IllegalStateException
     {
-        ResultPage<E> resultPage = mResultPages.get(mDirection);
-        switch (mDirection)
+        ResultPage<Envelope<Event>> resultPage = mLastResultPages.get(direction);
+        if (!hasComingPage(direction))
+        {
+            throw new IllegalStateException("No coming page for direction: " + direction);
+        }
+        switch (direction)
```
dmfs: no comment ;-)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/itemsprovider/EventListItemsProviderImpl.java
```diff
@@ -102,6 +98,8 @@ public EventListItemsProviderImpl(FutureServiceConnection<ApiService> apiService
         mIsLoading = new HashMap<>(2);
         mIsLoading.put(TOP, false);
         mIsLoading.put(BOTTOM, false);
```
dmfs, line 97: This line is when the HashMap gets reashed. Btw. there is an EnumMap in Java that optimized for this case (because the maximum number of keys is known and they don't need to be hashed).

dmfs (changes requested):

dmfs:
In general it appears to work, but sometimes I hit the end and I have to pull again to see the loading indicator.

lemonboston:
@dmfs thanks for the inputs, I made the changes, please review again.

The HashMap rehash for capacity 2, is that because of the default loadfactor 0.75? Anwyay, thanks for the tip on EnumMap, I forgot about its existence. I replaced them.

About `ScrollDirection`: I used these smartier enums earlier myself, too, but here I looked at it more like a parameter that's used widely at many different kind places rather than something that should handle things on its own. But the suggested changes clearly made the code simpler, so I made them. Even though it is now used both as parameter (like in `EdgeReachScrollListener` and `EventListItems`, and in the EnumMaps) and holds and does things on its own.

I've created a separate bug for the case you mentioned: https://github.com/schedjoules/android-event-discovery-sdk/issues/27

dmfs:
Yes, that capacity thing is related to the load factor. If 75% of the capacity is used, the map will be rehashed to make room. If the capacity is 2 that needs to happen when the 2nd element is about to be inserted. So when you create a HashMap and you can estimate the number of elements make sure the capacity you specify is as least about 33% larger than your estimated size.

dmfs (approved):

PR #22, opened 2016-12-08
Strings

dmfs:
@lemonboston please review.
I've changed a few strings, created different messages for when the bottom or top of the list has been reached and added translations.


lemonboston:
These are not used, can we delete them?:
```
<string name="schedjoules_entry_point">Find concerts</string>
<string name="schedjoules_app_name" translatable="false">SchedJoules Event Discovery</string>
<string name="schedjoules_event_details_no_description">No description</string>
``` 

This can be deleted from the NL version:
```
<string name="schedjoules_powered_by_footer" translatable="false">Powered by SchedJoules</string>
```

(The Translations Editor in Studio gives a nice overview by the way, marking items red which have issue. But it's seems also buggy as it shows one item missing in spite that it is not.)

Otherwise looks good.

dmfs:
We won't delete the entry point string. It's meant to be used as a menu option by partners. The others can certainly be removed.

dmfs:
Fixes applied.

PR #26, opened 2016-12-08
Fix Activity leak in EventListItemsProvider by using setters for depe…

### LINKED ISSUE #24: Activity is leaking on Event list screen rotation
Some recent refactors introduced an Activity leak by providing UI elements to a retained object here:
```
mListItemsProvider = retainedObjects.getOr(0,
                new EventListItemsProviderImpl(mApiService, new EventListItemsImpl(),
                        new EventListBackgroundMessage(this), new EventListLoadingIndicatorOverlay(this)));
```
The `this` keywords (activity) passing into a retained object shows this is not good.

We need to put back the earlier callback mechanism probably to notify the UI.
Worth also looking through how the RecyclerView and Adapter are retained/recreated.

lemonboston:
…ndencies that hold to Activity.

For #24.

Checked with Android Studio's Leaked Activities detection as well (before was positive, after negative).

@dmfs please review


dmfs:
please resolve the merge conflicts before I review.

lemonboston:
@dmfs updated with master, ready for review.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/EventListActivity.java
```diff
@@ -80,20 +79,20 @@ protected void onCreate(Bundle savedInstanceState)
...
                 new Intent("com.schedjoules.API").setPackage(getPackageName()));
 
-        mEventListItems = new EventListItemsImpl();
-        mListItemsProvider = retainedObjects.getOr(0,
-                new EventListItemsProviderImpl(mApiService, mEventListItems,
-                        new EventListBackgroundMessage(this), new EventListLoadingIndicatorOverlay(this)));
+        RetainedObjects retainedObjects = new Restoring(getLastCustomNonConfigurationInstance());
+        mListItemsProvider = retainedObjects.getOr(0, new EventListItemsProviderImpl(new EventListItemsImpl()));
+
+        mListItemsProvider.setApiService(mApiService);
```
dmfs, line 88: I can see why you need these setters now, but I don't like the solution. In general I'm curious if we can't just create a `Parcelable` state that's retained in a Bundle and used to create a new instance when the activity is resumed.
I'm doing something similar in the oauth2-essentials [`AuthorizationCodeGrant`](https://github.com/dmfs/oauth2-essentials/blob/master/src/main/java/org/dmfs/oauth2/client/grants/AuthorizationCodeGrant.java)

In some cases it might be required to pass the `AuthorizationCodeGrant` to another component using some sort of serialization (or even just to retain it during configuration changes in an activity). That's a problem because it needs to know the complex `OAuth2Client` object (which might need a context to be created). To overcome this it has a method `state()` that returns a `Serializable` factory object. After de-serialization this factory can be used to recreate the original `AuthorizationCodeGrant`. Essentially it works like this
```
AuthorizationCodeGrant authCodeGrant = new AuthorizationCodeGrant(client, scope, randomstateparam);
// do something with authCodeGrant

// need to create a serializable state
OAuth2InteractiveGrant.OAuth2GrantState state = authCodeGrant.state();
// serialize state

// after de-serializing "state" you can use it to recreate the AuthorizationCodeGrant
AuthorizationCodeGrant sameAuthCodeGrant = state.grant(client);
```
That's certainly a cleaner way.

The current method using `Restoring` is error prone, not only because you can easily leak the context. Also, using integers to refer to the retained data is easy to get wrong.

That being said, I see that we need to retain the `ResultPage` or the `ApiQuery` for the next page using the custom non configuration parameters, because these can not be serialized or parcelled (yet). But we should retain as few (and small) objects as possible that way.

lemonboston: Just some more thoughts on this:

After looking at current solution again and comparing to retained Fragment, I think now that that would be better, so if we need to touch this part again, I think it might be worth moving to Fragment. The reason is that in the Fragment all the objects that are needed to be retained would be initialized in onCreate(), and the view and other context specific ones in onCreateView(). Basically no extra magic needed as now.

About what needs to be retained now - isn't it the event list the main reason why don't want to do serialization now? I am not sure but wouldn't Serializing (/parcelling) possibly hundreds of events take noticeable time?
And currently all the event list loading mechanism and all the related state is deliberately contained in this `EventListItemProviderImpl`. So if we have to retain any of that state, it looks practical to just retain the whole object.

I checked out the `AuthorizationCodeGrant` state handling, I see the point there, looks like an interesting pattern, so I will keep it in mind.

lemonboston:
@dmfs do you request changes or are you going to merge it?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/EventListActivity.java
dmfs: I agree that it's probably not a good idea to bundle hundreds of events, though I don't think that it's a problem on normal configuration changes. `Bundle`s are quite lazy in that regard and they only serialze/deserialize when necessary. I believe this is not happening during a normal configuration change.
We should come up with a better way to cache the results of a query through configuration changes. If we can avoid this retention method (including retained fragments) we probably can save us a lot of trouble.
For now, please move it to a (retained) Fragment, because, as said before, it should be in a Fragment anyway.
I'll check if I can find a good way to serialize `ApiQuery`s, so we can save the previous and next page queries easily.

lemonboston:
Ok, I will do the Fragment refactor, but could we please merge this PR still to master first, so at least we have a fix already? These changes are an improvement already on the current master state.
The bug could even be left open until the new PR for the Fragment refactor is merged, although it would be effectively fixed by this PR, so more proper would be a 'technical-enhancement' task for the Fragment refactor.

dmfs:
There is no point in merging this to master because the changes of this are mostly obsolete when we move to a fragment solution. I'd prefer if you start to work on the fragment solution right away, because it might affect some of the following tickets. So having this fixed first can save us form a number of merge conflicts.

dmfs:
superseded by #36

PR #28, opened 2016-12-08
Add capability to launch event details with just the UID.

dmfs:
Still needs a few improvements though.
Fixes a few strings.
@lemonboston please do an initial review.


lemonboston:
I've mostly looked it through, here are my initial thoughts, comments:

```
new UidEventDetails("ada36d012294c9c4").show(this);
```
I like this declarative launching, too. I think it could possibly be used for almost any kind of UI component, like showing Fragments as well, or dialog, toast, etc. So current interface `EventDetails` could be more general. Alternative names for `show()` might be `launch()`, `open()`.

I think it would be great to handle the other side elegantly as well, so extracting from intents. Maybe something like:
```
EventDetailIntent detailIntent = new EventDetailIntent(getIntent());
if (detailIntent.isByUid()) {
   ... detailIntent.eventUid()...
   //..
}
else if (detailIntent.isByEvent())
{
   Event event = detailIntent.event();
  //..
}
```
It wouldn't be an `android.Intent` of course, but I can't think of a better name, '*Request' would be ambiguous, and it is also related to the actual intent, so I think it would be fine. This would be a special class for each screen, no generalization needed I think. Main benefit would be to have cleaner code in the activity, for example in the current state of EventDetailsActivity.onCreate(), it's not clear what is happening from the first look.
And I would move all our intent action and extra constants to a single class, to not litter the Activities and other classes, since they are now sort of hidden behind the declarative classes. And also to enable simpler reuse of them.
What do you think?

Regarding Service, Jobs, etc:
I see the usefulness of `FutureServiceConnection` and `AbstractApiService` much better now.
And I see that it might be useful to create something like the `ServiceJobQueue` and `ServiceJob`.
But I realized after a while that the jobqueue and the executorservice might need to be in the actual `android.app.Service`. Because their lifetime should probably be the same.
I know you deliberately haven't introduced async for the api-client and maybe for the ApiService, too, but we could consider if the app would use the queue always, then make it sort of default. But even if we want to allow to let it decoupled, so we can use different solutions in different components, then still it could make sense I think to make the `ServiceJobQueue` own the connection, so sort of putting a layer on top of the connection. So I am thinking in something like this:
```
public interface ServiceJobQueue<T>
{
    void post(ServiceJob job);

    void shutdown();
}
```
It would create connection and thread pool in constructor and disconnect, shutdown in shutdown().
Currently JobQueue disconnects after each job, it looks like a bug now - there is no reconnection.
I think we need to keep connecting, disconnecting exposed at android lifecycle level, so the top component we use. So either manage both FutureSericeConnection and ServiceJobQueue or sort of merge them as suggested above.

About the current `ServiceJob`: we need to manually handle main thread delivery now, and `showError()` is actually called both from background thread currently, so those seem like bugs.
So I think it would be good to have main thread callback, so basically something very similar to the AsyncTasks we have. Something like:
```
public interface ServiceJob<S, T>
{    
    @WorkerThread
    T execute(S service);

    @MainThread
     void onResult(ThrowingResult<T> result);
}
```
Not sure what is the best way: to use the existing `AsyncTask` (maybe even just under the hood) or just implement our own handling in the jobqueue. We can easily post to main thread with `new Handler(Looper.getMainLooper()).post(..)`, but I am not sure what else is there in `AsyncTask` that might be useful.
Or is your intent with `ServiceJobQueue` still to be a lower level component, so not handling main thread? I think if it's a place of threading anyway, it can be smarter and be able to deliver on main thread.

About the executor, I have a concern using `AsyncTask.SERIAL_EXECUTOR` for networking because it's a global serial executor, might be used elsewhere, so anything else could block it. I think it might be better to decouple from it and use a dedicated executor by default.

Why is `ServiceJob.onTimeOut()` needed if there is a try catch anyway in `execute()`?

`ApiService.FutureConnection` - nice that it makes usage simpler. The class in the interface doesnt' look very nice for me though, could it be a top level class? (eg: ApiFutureServiceConnection)

`executionTimeout` param in `ServiceJobQueue.post(..)` - is it need per job? Could it be constructor param to the JobQueue?

Binding to application context in `FutureServiceConnection` - shouldn't we consider using the actual contexts (activity) instead? Just in case we forgot to disconnect() - which we should not of course, but in that case, Service would be kept bound to the partner application even if event discovery is not shown. With activity, it is disconnected when that is destroyed as I know. So I don't see drawback, just benefit of using the actual context. Do I miss something?

`schedjoules_fragment_error.xml`
Maybe rename to `schedjoules_fragment_event_detail_error.xml`?
Some copy-paste leftover:
```
android:orientation="vertical"
android:animateLayoutChanges="true"
```
Why do you prefer these general ids? It's almost impossible to back track to them from code. Isn't there a risk that in a big layout these general ids would clash?
```
android:id="@android:id/message"
```
Shouldn't it be at least `schedjoules_message`?
I see you used the `layout` root, it's probably no benefit here, though.
Especially if the id is android:id as I just checked in `schedjoules_fragment_loader.xml`, the binding doesn't generate fields for that.

Bug currently: event title is not shown on the toolbar

Also, the content is shown abruptly after the loading currently.

Maybe a less specific name for this?
```
<string name="schedjoules_this_might_take_a_moment">This might take a moment</string>
```

dmfs:
Check out this branch:
https://github.com/schedjoules/android-event-discovery-sdk/tree/stories/25-details-by-uid-wizard
There is still a lot of room for improvement and it's more a proof of
concept than polished code.
The idea is to have multiple smaller fragments with a very special
purpose (load event, present event, present error) and transition
between these. The transitions are controlled by the fragments, not by
the activity. The activity merely hosts the "wizard". Fragments can
start the transition to another Fragment by starting another WizardStep.
That stuff is certainly not perfect yet, but it already works well. The
main advantage is we don't have a single complex fragment, but multiple
rather small (possibly recyclable) ones.

Also see my comments below.

> I've mostly looked it through, here are my initial thoughts, comments:
>
> |new UidEventDetails("ada36d012294c9c4").show(this); |
>
> I like this declarative launching, too. I think it could possibly be
> used for almost any kind of UI component, like showing Fragments as
> well, or dialog, toast, etc. So current interface |EventDetails| could
> be more general. Alternative names for |show()| might be |launch()|,
> |open()|.
>
The name should express the intent. In this case I want to `show` the
`EventDetails`, hence the name. A general interface isn't of much use,
because there will be other "builder"-style methods that usually depend
on the subject.
>
> I think it would be great to handle the other side elegantly as well,
> so extracting from intents. Maybe something like:
>
> |EventDetailIntent detailIntent = new EventDetailIntent(getIntent());
> if (detailIntent.isByUid()) { ... detailIntent.eventUid()... //.. }
> else if (detailIntent.isByEvent()) { Event event =
> detailIntent.event(); //.. } |
>
> It wouldn't be an |android.Intent| of course, but I can't think of a
> better name, '*Request' would be ambiguous, and it is also related to
> the actual intent, so I think it would be fine. This would be a
> special class for each screen, no generalization needed I think. Main
> benefit would be to have cleaner code in the activity, for example in
> the current state of EventDetailsActivity.onCreate(), it's not clear
> what is happening from the first look.
> And I would move all our intent action and extra constants to a single
> class, to not litter the Activities and other classes, since they are
> now sort of hidden behind the declarative classes. And also to enable
> simpler reuse of them.
> What do you think?
>
Yeah, I need to think about this. In general I like the idea.
>
> Regarding Service, Jobs, etc:
> I see the usefulness of |FutureServiceConnection| and
> |AbstractApiService| much better now.
> And I see that it might be useful to create something like the
> |ServiceJobQueue| and |ServiceJob|.
> But I realized after a while that the jobqueue and the executorservice
> might need to be in the actual |android.app.Service|. Because their
> lifetime should probably be the same.
> I know you deliberately haven't introduced async for the api-client
> and maybe for the ApiService, too, but we could consider if the app
> would use the queue always, then make it sort of default. But even if
> we want to allow to let it decoupled, so we can use different
> solutions in different components, then still it could make sense I
> think to make the |ServiceJobQueue| own the connection, so sort of
> putting a layer on top of the connection. So I am thinking in
> something like this:
>
> |public interface ServiceJobQueue<T> { void post(ServiceJob job); void
> shutdown(); } |
>
> It would create connection and thread pool in constructor and
> disconnect, shutdown in shutdown().
> Currently JobQueue disconnects after each job, it looks like a bug now
> - there is no reconnection.
>
calling `service` would reconnect automatically. However could cause a
race condition, so I've changed it (in the wizard branch).
>
> I think we need to keep connecting, disconnecting exposed at android
> lifecycle level, so the top component we use. So either manage both
> FutureSericeConnection and ServiceJobQueue or sort of merge them as
> suggested above.
>
I've implemented a `disconnect` method as suggested.
>
> About the current |ServiceJob|: we need to manually handle main thread
> delivery now, and |showError()| is actually called both from
> background thread currently, so those seem like bugs.
> So I think it would be good to have main thread callback, so basically
> something very similar to the AsyncTasks we have. Something like:
>
> |public interface ServiceJob<S, T> { @WorkerThread T execute(S
> service); @MainThread void onResult(ThrowingResult<T> result); } |
>
> Not sure what is the best way: to use the existing |AsyncTask| (maybe
> even just under the hood) or just implement our own handling in the
> jobqueue. We can easily post to main thread with |new
> Handler(Looper.getMainLooper()).post(..)|, but I am not sure what else
> is there in |AsyncTask| that might be useful.
> Or is your intent with |ServiceJobQueue| still to be a lower level
> component, so not handling main thread? I think if it's a place of
> threading anyway, it can be smarter and be able to deliver on main thread.
>
I think in a addition to ServiceJob we need a second interface.
ServiceJob is not really designed to deliver a result. It's meant to
"post" something to a service without actually returning any result. To
return the result, an extended mechanism would be useful.
>
> About the executor, I have a concern using |AsyncTask.SERIAL_EXECUTOR|
> for networking because it's a global serial executor, might be used
> elsewhere, so anything else could block it. I think it might be better
> to decouple from it and use a dedicated executor by default.
>
Good point.
>
> Why is |ServiceJob.onTimeOut()| needed if there is a try catch anyway
> in |execute()|?
>
This is about the connection timeout to the service. If the JobQueue
can't connect to the service `execute()` is never called, but we still
need to inform the Job about this error. So this is not about network
timeouts.
>
> |ApiService.FutureConnection| - nice that it makes usage simpler. The
> class in the interface doesnt' look very nice for me though, could it
> be a top level class? (eg: ApiFutureServiceConnection)
>
I wanted to make it an inner class to express the close relationship.
Making it a top-level class would kind of disconnect them.
>
> |executionTimeout| param in |ServiceJobQueue.post(..)| - is it need
> per job? Could it be constructor param to the JobQueue?
>
Some Jobs can wait longer for execution than others. Not sure if that
actually makes sense, but I felt this is something job-specific not
queue-specific.
>
> Binding to application context in |FutureServiceConnection| -
> shouldn't we consider using the actual contexts (activity) instead?
> Just in case we forgot to disconnect() - which we should not of
> course, but in that case, Service would be kept bound to the partner
> application even if event discovery is not shown. With activity, it is
> disconnected when that is destroyed as I know. So I don't see
> drawback, just benefit of using the actual context. Do I miss something?
>
I wanted to avoid leaking the context if the caller forgets to unbind.
Not sure if that's possible after all though. I need to check that.
>
> |schedjoules_fragment_error.xml|
> Maybe rename to |schedjoules_fragment_event_detail_error.xml|?
> Some copy-paste leftover:
>
> |android:orientation="vertical" android:animateLayoutChanges="true" |
>
> Why do you prefer these general ids? It's almost impossible to back
> track to them from code. Isn't there a risk that in a big layout these
> general ids would clash?
>
> |android:id="@android:id/message" |
>
> Shouldn't it be at least |schedjoules_message|?
>
I don't have any specific reason. Maybe it's because I'm too lazy to
think of a name ;-).
>
> I see you used the |layout| root, it's probably no benefit here, though.
> Especially if the id is android:id as I just checked in
> |schedjoules_fragment_loader.xml|, the binding doesn't generate fields
> for that.
>
That was just a result of copy & paste. I didn't bother to remove
everything I don't need.
>
> Bug currently: event title is not shown on the toolbar
>
Yep, that's known and should already be fixed.
>
> Also, the content is shown abruptly after the loading currently.
>
Should be fixed in the "wizard" branch
>
> Maybe a less specific name for this?
>
> |<string name="schedjoules_this_might_take_a_moment">This might take a
> moment</string> |

dmfs:
This has been superseded by #34, hence I'm closing this.

PR #32, opened 2016-12-13
PR for 17 date header format

### LINKED ISSUE #17: Broken date header date format
The date header format is broken. In particular it's not properly localized.
Please use DateUtils to generate a short format with day of week, day of month and month.
If the date is not in the same year as today, the year should be present too.

DateUtils also supports a relative date format that automatically takes care of "today" and "tomorrow".

lemonboston:
Resolves #17 

@dmfs please review.
Please check the formats for the different locales as well in `SmartFormattedDayAndroidTest` if they look good or should we use full day or month texts instead of the abbreviations.

Also, should we maybe add the date after "Today" and "Tomorrow" in parenthesis? Like:
Today (Wed, 11 Oct)
I am not sure if that's better, so just asking.


lemonboston:
sorry, hold on a little, I have to check a test again..

lemonboston:
Fixed it and now the tests actually test the displayed text.

Wanted to mention also that I added this general interface for the formatted texts:
```
public interface FormattedDateTime
{
     CharSequence value(Context context);
}
```
It's not a `CharSequence` in itself, because for that we would need to pass in `Context` in the constructor and then either we keep a reference, which would risk leaking, although we could take `context.getApplicationContext()`, or create the string eagerly there in constructor, which is again not nice. So I thought having context as parameter in the method that actually creates the text feels better, it's like we have a promise of the text with the object but to actually create it we need an 'execution' parameter, the `Context`. Which is kind of right, isn't it?
What do you think?
This also allows to create the promise of the text in advance as in current `DateHeaderItem` constructor, so it is already useful there.

(@dmfs ready for review)

lemonboston:
Which format should we use?
Both the day of the week name and the month name can be full or abbreviated independently.
 
short-short-en
long-long-en
short-short-de
long-long-de

I don't have Dutch on my phone, but the short ones look like this:
di 25 okt.
di 18 okt.

dmfs:
An the winner is: short day, long month

lemonboston:
@dmfs updated, please reivew.

dmfs (approved):

PR #34, opened 2016-12-14
Stories/25 details by uid wizard

dmfs:


lemonboston:
I looked into it, not every part, just to get the idea about the new components and changes. 

Notes/questions:

---
Why is starting the bounded Services needed? For the in memory cache to keep it there?
Isn’t it not nice to let it live after we don’t use it? I know it would probably be shut down by the system after a while, but still.
Couldn't we use disk backed cache to avoid this?
(DiskLruCache-s: [link1](https://developer.android.com/samples/DisplayingBitmaps/src/com.example.android.displayingbitmaps/util/DiskLruCache.html) [link2](https://github.com/JakeWharton/DiskLruCache))

Would the Http cache be better?
Shouldn’t caching be transparent? It adds to the complexity now to make cache call separately, or was it deliberate to make it possible to get it before the first `WizardStep`?

----
```
BasicActionsService

@Override
public final void onCreate()
{
   super.onCreate();
   mApiServiceConnection = new ApiService.FutureConnection(this);
}
```
Doesn't this bind the Service to itself? Not sure why would Andriod even allow that, but wouldn’t it mean it will never stop?

---
Souldn’t disconnect() handle case when it’s not connected? (hiding this `if` in `FutureLocalServiceConnection` if needed at all)
```
BasicEventDetails

finally
{
   if (actionService.isConnected())
   {
       actionService.disconnect();
   }
}
```
---
The boilerplates around intent and bundle extras handling is still bothering me, I think it makes logic harder to read, adds code duplication and also makes it error prone. Not sure what would be a really nice approach to hide them, though. So just a quick idea about the nested bundles:
```
new PutNestedBundle(string, parcelable).into(intent)
T obj = new TakeNestedParcelable(string).from(intent)
```
Maybe for specific parameters as well, having two inner class for put and get:
```
new EventExtra.Put(event).into(intent)
..
Event event = new EventExtra.Get().from(getIntent())
```
So the extra key constants, the parcelable handling (decoration, casting), all could be hidden in `EventExtra`. And of course there would be an interface for these extras/args, not sure how good that works with that inner classes.

---
`schedjoules_event_detail_content_loading_actions.xml`
`android:id="@+id/header"`
SchedJoules prefix?

---
`ActionLoaderStep.LoaderFragment`
`private LinearLayout mVerticalItems;` not used → can be removed from layout

lemonboston (commented):
(comment in the conversation)

dmfs:
> Why is starting the bounded Services needed? For the in memory cache
> to keep it there?
Two reasons:

* To reduce latency when we actually bind it the first time
* To reduce the chances of it being stopped when we unbind, again to
avoid startup overhead causing latency when we need it again, but also
to retain the cache.
>
> Isn’t it not nice to let it live after we don’t use it? I know it
> would probably be shut down by the system after a while, but still.
>
I don't consider this an issue, especially since that's a service that
we frequently use. The nice thing about this solution is that we don't
have to maintain a single connection and pass it to every component that
needs it. A component can create it's own connection and disconnect as
long as it has a context. If the service is already bound, binding it
again is very cheap and has no notable overhead.
>
> Couldn't we use disk backed cache to avoid this?
> (DiskLruCache-s: link1
> <https://developer.android.com/samples/DisplayingBitmaps/src/com.example.android.displayingbitmaps/util/DiskLruCache.html>
> link2 <https://github.com/JakeWharton/DiskLruCache>)
>
I don't like that DiskLruCache implementation. The design is pretty bad.
>
> Would the Http cache be better?
>
Yes, an http cache is an important component. But it's too complex to be
implemented quickly. The current solution is sort of a quick win.
Another important feature is the ability to check if an item is in the
cache without fetching it if it's not. That's more difficult with an
HTTP cache (because it usually performs a network request if the cache
doesn't deliver the response). So we'll probably keep this "seesion
cache" even when an http cache is in place.
The overall goal of this is to reduce UI latency. So with the new
changes you won't see any delay or loader animation when you re-open an
event of which the actions are already cached.
>
> Shouldn’t caching be transparent? It adds to the complexity now to
> make cache call separately, or was it deliberate to make it possible
> to get it before the first |WizardStep|?
>
Exactly. I needed a way to get any cached item quickly without
triggering a network request in case of a cache miss. In case of a cache
miss I'm just launching a different WizardStep (one that fetches the
missing item from the API) but in case of a cache hit I can launch the
ShowEventStep directly.
> ------------------------------------------------------------------------
> |BasicActionsService @Override public final void onCreate() {
> super.onCreate(); mApiServiceConnection = new
> ApiService.FutureConnection(this); } |
>
> Doesn't this bind the Service to itself? Not sure why would Andriod
> even allow that, but wouldn’t it mean it will never stop?
>
Yes, this triggers the bind, but it doesn't wait for the service to be
bound and returns instantly. So it doesn't block.
> ------------------------------------------------------------------------
>
> Souldn’t disconnect() handle case when it’s not connected? (hiding
> this |if| in |FutureLocalServiceConnection| if needed at all)
>
> |BasicEventDetails finally { if (actionService.isConnected()) {
> actionService.disconnect(); } }|
|That's probably a good idea. I'll check that. I also need to look
deeper into the service binding mechanism. There are still a few open
questions.|
> ||
> ------------------------------------------------------------------------
>
> The boilerplates around intent and bundle extras handling is still
> bothering me, I think it makes logic harder to read, adds code
> duplication and also makes it error prone. Not sure what would be a
> really nice approach to hide them, though. So just a quick idea about
> the nested bundles:
>
> |new PutNestedBundle(string, parcelable).into(intent) T obj = new
> TakeNestedParcelable(string).from(intent) |
>
> Maybe for specific parameters as well, having two inner class for put
> and get:
>
> |new EventExtra.Put(event).into(intent) .. Event event = new
> EventExtra.Get().from(getIntent()) |
>
> So the extra key constants, the parcelable handling (decoration,
> casting), all could be hidden in |EventExtra|. And of course there
> would be an interface for these extras/args, not sure how good that
> works with that inner classes.
>
It's bothering me too, but the proposed code is not an option. Put and
Get don't sound like proper classes. If anything we probably should have
a wrapper for the intent itself. Similar to the "launcher builders".
> ------------------------------------------------------------------------
>
> |schedjoules_event_detail_content_loading_actions.xml|
> |android:id="@+id/header"|
> SchedJoules prefix?
>
Ah, yes.
> ------------------------------------------------------------------------
>
> |ActionLoaderStep.LoaderFragment|
> |private LinearLayout mVerticalItems;| not used → can be removed from
> layout
>
will do.

lemonboston:
Thanks for the reply.

```
BasicActionsService

@Override
public final void onCreate()
{
   super.onCreate();
   mApiServiceConnection = new ApiService.FutureConnection(this);
}
```
>> Doesn't this bind the Service to itself? Not sure why would Andriod
 even allow that, but wouldn’t it mean it will never stop?

> Yes, this triggers the bind, but it doesn't wait for the service to be
bound and returns instantly. So it doesn't block.

I wanted to say here that the Service binds to itself, so the Context/Component given to which to bind the Service is the Service itself. Isn't that a problem? Bounded Service can unbound itself and stop automatically if the component is destroyed if I know correctly, but this way there is a kind of circular relation, so it would never be destroyed. (?)

lemonboston:
Sorry, missed it totally as you mentioned that ActionService binds to ApiService, not to itself, so there is no problem there.

dmfs:
Great I'll merge then.

PR #36, opened 2016-12-15
PR for 24 - Refactor EventListActivity into Fragment

lemonboston:
#24 

@dmfs please review.

One issue come up with Glide, it's probably not important:
Glide loading can be initiated with Glide.with(context), Glide.with(activity) or Glide.with(fragment), and it uses the given component to pause and restart loading on lifecycle callbacks. So far we used Glide.with(context) that actually delegated to Glide.with(activity), but after the refactor, there was a crash on rotation saying that the activity is destroyed already. We should pass in the fragment, but that's not obvious how to access it there (EventItemView). So I use application context now, which means pause, restart on lifecyles won't happen, but I think it's fine. Let me know if you see any problem with it.
(Glide wiki mentions GIFs as a use case for the pausing, in which case it's clearly useful, but for our small thumbnails, I guess it's fine if they load possibly 'in vain' if the screen is recreated or left.)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/EventListFragment.java
```diff
@@ -0,0 +1,243 @@
...
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState)
+    {
+        super.onCreate(savedInstanceState);
+        setRetainInstance(true);
+        mIsInitializing = true;
+
+        // Application context used now, if changed to Activity, this and disconnect() has to move to other callbacks
+        mApiService = new FutureLocalServiceConnection<>(getContext().getApplicationContext(),
```
dmfs, line 97: Please use the new method to create the service connection
```
   mApiService = ApiService.FutureConnection(getActivity()).
```
FutureConnection uses application context internally. This is one of the reasons for that.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListToolbar.java
```diff
@@ -0,0 +1,86 @@
...
+
+
+    public EventListToolbar(Toolbar toolbar, Listener listener)
+    {
+        mToolbar = toolbar;
+        mListener = listener;
+    }
+
+
+    public void initToolbar(FragmentActivity fragmentActivity)
```
dmfs, line 47: Which interface does this implement?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListToolbar.java
```diff
@@ -0,0 +1,86 @@
...
+            activity.getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+        }
+        else
+        {
+            mToolbar.setTitleMarginStart(res.getDimensionPixelSize(R.dimen.schedjoules_list_item_horizontal_margin));
+        }
+    }
+
+
+    public void setToolbarTitle(CharSequence title)
```
dmfs, line 75: interface?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListView.java
```diff
@@ -0,0 +1,63 @@
...
+import eu.davidea.flexibleadapter.FlexibleAdapter;
+import eu.davidea.flexibleadapter.items.IFlexible;
+
+
+/**
+ * Handles the RecyclerView and the Adapter initialization for the Event list.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class EventListView
```
dmfs, line 33: interface?

dmfs (changes requested):
Considering the scope of this story (move existing code form activity to fragment), this is okay, apart from a few minor changes.
However in general I disagree with parts of the architecture. My gut feeling says the way the components are structured is not right yet.
We should take care of that in a separate story.

dmfs:
Merging this for now to get it into today's release. Please don't consider this completed.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/EventListFragment.java
lemonboston: Updated

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListToolbar.java
lemonboston: With `EventListToolbar` and `EventListView` (this name may not be the best) the motivation was to create an object kind of declaratively for a UI item, so it can contain and hide its corresponding boilerplate and initialization code. It is very specific to the actual UI component on the actual screen, so that's why I didn't create an interface. It would feel strange for me to create an interface for such a specific and simple classes. The only possible general method I can think of might be `init(Activity)`.
I usually only create interface for these kind of specific classes when they grow bigger, to have an overview what they do by looking at the interface.
But of course I can create the interfaces, let me know if I should, or if you have any other idea in mind how to handle this.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListToolbar.java
lemonboston: see above

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListView.java
lemonboston: see above

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListToolbar.java
dmfs: Yes, the rule is every public method implements an interface. No Exceptions (at least not for our own classes).
Also, why don't you pass the Activity to the constructor? You already pass a View, which means you're already holding a context anyway.
Please make sure the design is clean.

lemonboston: continuing this in new PR:
https://github.com/schedjoules/android-event-discovery-sdk/pull/45

PR #38, opened 2016-12-16
PR for 16 - Force chooser upon share

lemonboston:
@dmfs please review

I've added the following string as the title of the chooser, DE, NL versions could be added:
<string name="schedjoules_share_title">Share Event information</string>

There is a possibility from API level 22 to get a callback on what package the user selected in the chooser. Should we add that as Insight in a separate story?
(Descriptions:
https://developer.android.com/reference/android/content/Intent.html#createChooser(android.content.Intent, java.lang.CharSequence, android.content.IntentSender)
http://stackoverflow.com/questions/30631902/get-intentsender-object-for-createchooser-method-in-android/30693465#30693465 )

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/actions/ShareActionExecutable.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/actions/ShareActionExecutable.java
@@ -17,9 +17,11 @@
 
 package com.schedjoules.eventdiscovery.actions;
 
+import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.support.annotation.NonNull;
+import android.support.v4.app.ShareCompat;
 import android.widget.Toast;
 
 import com.schedjoules.client.eventsdiscovery.Event;
@@ -60,10 +62,24 @@ public final class ShareActionExecutable implements ActionExecutable
 
         String eventText = eventText(context);
 
-        Intent sendIntent = new Intent();
-        sendIntent.setAction(Intent.ACTION_SEND);
-        sendIntent.putExtra(Intent.EXTRA_TEXT, eventText);
-        sendIntent.setType("text/plain");
+        Intent sendIntent;
+        if (context instanceof Activity)
+        {
+            sendIntent = ShareCompat.IntentBuilder
+                    .from((Activity) context)
+                    .setText(eventText)
+                    .setType("text/plain")
+                    .setChooserTitle(R.string.schedjoules_share_title)
+                    .createChooserIntent();
+        }
+        else
+        {
+            Intent pureIntent = new Intent();
+            pureIntent.setAction(Intent.ACTION_SEND);
+            pureIntent.putExtra(Intent.EXTRA_TEXT, eventText);
+            pureIntent.setType("text/plain");
+            sendIntent = Intent.createChooser(pureIntent, context.getString(R.string.schedjoules_share_title));
+        }
 
         if (sendIntent.resolveActivity(context.getPackageManager()) != null)
         {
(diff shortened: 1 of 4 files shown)


dmfs:
Yes that sounds like an interesting insight. Please add two stories for that (one in java-api-client and one here).

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/actions/ShareActionExecutable.java
```diff
@@ -60,10 +62,24 @@ public void execute(@NonNull Context context)
 
         String eventText = eventText(context);
 
-        Intent sendIntent = new Intent();
-        sendIntent.setAction(Intent.ACTION_SEND);
-        sendIntent.putExtra(Intent.EXTRA_TEXT, eventText);
-        sendIntent.setType("text/plain");
+        Intent sendIntent;
+        if (context instanceof Activity)
```
dmfs, line 66: Why do we need this test? Can't we just pass an `Activity` instead of a `Context`? I can't imagine any use case for launching this from something that's not an `Activity`.

eventdiscovery-sdk/src/main/res/values/strings.xml
```diff
@@ -41,5 +41,6 @@
 
     <string name="schedjoules_title_error">Error</string>
     <string name="schedjoules_generic_error">Something went wrong.\nPlease try again later.</string>
+    <string name="schedjoules_share_title">Share Event information</string>
```
dmfs: "Share Event"  is sufficient.
in Dutch "Delen …"
in German "Ereignis teilen"

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/actions/ShareActionExecutable.java
lemonboston: I thought about action buttons on Notifications as a theoretical possibility.
For example getting directions could be a button on a notification.

eventdiscovery-sdk/src/main/res/values/strings.xml
lemonboston: Updated.

lemonboston:
Stories created for the share target insight.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/actions/ShareActionExecutable.java
dmfs: It's true that some actions might be called from notifications, but here we're talking about the Share Action, which I can guarantee will never be an option on a notification.

lemonboston: I though you meant by _"Can't we just pass an `Activity` instead of a `Context`?"_ that we change the `ActionExecutable.execute()` parameter to `Activity`...
So you mean just removing the if-else?
Note: the callback for the selected target only works with a call that's not available in `ShareCompat` as I see it. So we will need to use the version in current `else` branch when we add that.

dmfs: Yes, we should change the parameter to `Activity`. At present, I don't think that we'll use ActionExecutables from any other Context.

I think `ShareCompat.IntentBuilder` can return the "pureIntent" as well, so you still can use it.

dmfs:
I've merged it into today's release, but please don't consider this completed yet.

Issue #41, opened 2016-12-17
In order to be able to jump to right event in app, we need deeplinking in place


dmfs:
@rutgerg commented on [Mon Dec 12 2016](https://github.com/schedjoules/backlog/issues/81)

PR #42, opened 2016-12-17
Add support for deep links. Deep links are catched by a proxy activit…

dmfs:
…y that forwards the request to the regular details activity. Implements #41


dmfs:
Btw. I call this pattern "activity decorator". It certainly has some significant overhead compared to the regular Java decorator pattern, so you can't decorate more often than once or twice. However, I like how it allows you to add functionality without modifying the original Activity.

dmfs:
@lemonboston please review, although already merged

lemonboston (commented):
Looks good.

PR #45, opened 2016-12-20
PR for 24 cont

lemonboston:
continuing https://github.com/schedjoules/android-event-discovery-sdk/pull/36


lemonboston:
> Yes, the rule is every public method implements an interface. No Exceptions (at least not for our own classes).
> Also, why don't you pass the Activity to the constructor? You already pass a View, which means you're already holding a context anyway.
> Please make sure the design is clean.

@dmfs please review
I made the changes and it does get cleaner with the interfaces, even here.
I put initialization logic in constructors, which is usually not good, but it feels correct here for me, let me know if you would prefer an init() method instead.

lemonboston:
linking the story: #24

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/FlexibleAdapterEventListView.java
```diff
@@ -0,0 +1,58 @@
...
+import eu.davidea.flexibleadapter.FlexibleAdapter;
+import eu.davidea.flexibleadapter.items.IFlexible;
+
+
+/**
+ * Handles the RecyclerView and the Adapter initialization for the Event list.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class FlexibleAdapterEventListView implements EventListView
```
dmfs: What's the real "single purpose" if this class? The name says it's a "View". The description says it "handles" something. To me it looks like an alibi class for something that was meant to go into a static method.
This should be refactored. Do we really need this "class"?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListToolbarImpl.java
```diff
@@ -0,0 +1,75 @@
...
+import com.schedjoules.eventdiscovery.R;
+import com.schedjoules.eventdiscovery.common.BaseActivity;
+
+
+/**
+ * The toolbar of the Event list screen.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class EventListToolbarImpl implements EventListToolbar
```
dmfs: What's this class all about? What does it add? Could the initialization go to a method instead? The `setToolbarTitle` method doesn't seem to add anything.
Is this really a class?

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/FlexibleAdapterEventListView.java
lemonboston: I agree that it's a rather unfortunate class like this and even the responsibility is not clear.
The purpose of both `EventListView` and `EventListToolbar` is to extract the low level UI initialization (and possibly controlling) code from the Fragment. They serve as kind of the View or ViewController in an MVP like thinking, so they serve as an handler for the UI, similar as the ScreenView was in the first version of the Activity. With the Toolbar, it's more clear I think, we might even say that we declare a Toolbar, with the EventListView, because of the adapter, and also because many related loading logic is not in this class, it is not really good, I agree.

The question is what would be a better way? Leaving everything in the Fragment/Activity that we couldn't find a very good object representation of leads to build up of various low-level code there, growing the classes too big and making it harder to read those controller level logics.
So the question from the other side is what is the Fragment's responsibility? We know it's sort of corrupted by default, but it shouldn't mean we give up on it and pour everything in it. I think what they have to do as a minimum in our current usage is the initialization of the components and acting as presentation layer controller to handle the high level callbacks, both the lifecycles and the ones from our own components. I think it would be best if everything else is delegated, and it doesn't contain low-level code.
So if we move those code under discussion now back to the Fragment, then it will contain methods like this:
```
    @Override
    public void onLocationSelected(LocationSelectionResult result)
    {
        mLastSelectedLocation.update(result);
        mToolbar.setToolbarTitle(result.name());
        mListItemsProvider.loadEvents(result.geoLocation(), startAfter());
    }
```
```
private void setupToolabr()
    {
        mToolbar = toolbar;

        mToolbar.setTitle(""); // Need to set it to something here, otherwise it will use title from Activity label
        mToolbar.setOnClickListener(new View.OnClickListener()
        {
            @Override
            public void onClick(View v)
            {
                listener.onToolbarTitleClick();
            }
        });

        BaseActivity activity = (BaseActivity) fragmentActivity;
        activity.setSupportActionBar(mToolbar);

        Resources res = activity.getResources();
        if (res.getBoolean(R.bool.schedjoules_enableBackArrowOnEventListScreen))
        {
            //noinspection ConstantConditions
            activity.getSupportActionBar().setDisplayHomeAsUpEnabled(true);
        }
        else
        {
            mToolbar.setTitleMarginStart(res.getDimensionPixelSize(R.dimen.schedjoules_list_item_horizontal_margin));
        }
    }
```
They have clearly different level of abstractions, the first one is high-level, delegating, the second one low-level. Why would a high-level orchestrator component need to know and deal with what is the margin of the toolbar and what extra framework boilerplate is needed to wire things together?

So my main concern is that I think we should keep the Fragment clean as well, not just the objects we create and I don't see a better option currently, although there might be of course.
So I propose moving `EventListView` code back to Fragment but keeping `EventListToolbar` as an object that represents the Toolbar and hides its initiation logic.

Let me know how you would refactor this, I will do that.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/EventListToolbarImpl.java
lemonboston: See above.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventlist/view/FlexibleAdapterEventListView.java
dmfs: For now move it to the fragment until we have a clear pattern to solve this. It's true that the current code hides some of the complexity by introducing another abstraction layer, in a way. However, in this particular case I don't consider this an advantage, because it also hides the fact that there is something wrong with the architecture and layering that needs to be fixed. I prefer to make it more obvious that we still have something to do in this regard.

lemonboston: Ok, done.

lemonboston:
@dmfs Can this be merged?

PR #56, opened 2016-12-22
PR for 21 - UI tests setup, Mock Api, Tests for the event list

lemonboston:
@dmfs can you please make an initial review?

Many more automated/manual test cases could be added, much of the effort so far gone into setting up the 'framework' around these tests.
So I plan to continue but it would also be good to have a review and possibly a merge at this point.


lemonboston:
@dmfs updated with master, please review.

lemonboston:
@dmfs Please review.
This is intended to be a first step towards creating a more extended UI test-suite for the SDK, something that already works and tests a few things, and which we can build/improve upon later.
The same true for the mocked Event ResultPages, I just progressed with that as the customization needs for the test arose.

eventdiscovery-sdk/src/androidTest/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListManualAndroidTest.java
```diff
@@ -0,0 +1,87 @@
...
+                                success(3), success(2), error(), error(), error(), success(3), empty()))
+                .setEventsOnPagesTop(
+                        asList(
+                                empty()))
+                .setPageLoadTime(1000)
+                .createMockPagesSetup());
+
+        new BasicEventDiscovery().start(mActivityTestRule.getActivity());
+
+        waitForMinutes(5);
```
dmfs, line 86: Why do you wait 5 minutes here?

lemonboston: It's just a manual test, meaning that we can setup some configuration for the API responses and start the app and do stuff ourselves. 5 minute wait is set so we have time to play around with the app, otherwise when the test ends, it closes the app.
But it is probably better if this test is not enabled by default, it doesn't make sense to run this as part of any build, so I disabled it by adding a `@Suppress` annotation to the class (`@Ignore` doesn't work for this Android test it seems.)

dmfs (approved):

PR #62, opened 2016-12-30
PR for 61 - Move feedback to menu

lemonboston:
@dmfs Please review. The notes below doesn't affect this PR.

I had some ideas about how to handle the Menu better and started to make the refactor, but I don't want to hold this up with that because that might need some rounds of discussion and review as well. So I would do that in separate PR if we get to that.
But that would be something like this:
```
public final class FeedbackMenuItem implements OptionsMenuItem
{
    @Override
    public CharSequence title(Context context)
    {
        return context.getString(R.string.schedjoules_menu_feedback);
    }


    @Override
    public void configure(MenuItem menuItem)
    {
        menuItem.setShowAsAction(MenuItem.SHOW_AS_ACTION_NEVER);
    }


    @Override
    public void onClick(Activity activity)
    {
        new ExternalUrlFeedbackForm().show(activity);
    }
}
```
So in the Fragment/Activity, it would be something like:
```
mMenu = new Compose(new LocationMenuItem(/*callback*/ this), new FeedbackMenuItem());
// + delegation in onCreateOptionsMenu() and onOptionsItemSelected();

// or even hide those delegations in BaseFragment and just use this single line in the Fragment:
addMenuItems(new LocationMenuItem(/*callback*/ this), new FeedbackMenuItem());
```
The UP button handling kind of relates to this as well because of the onOptionsItemSelected() callback - that could be set generally as well.
Also thinking if something like this would make sense:
```
interface OnClickAction
{
    void execute(Activity activity);
}
```
This could be used in place of the ActionExecutables, for these menu items, and for the UP button, and maybe at other places, too. It looks useful but the question is I guess, is it a problem that it's that general, and does it lead to more functional than oo design.


dmfs:
Please resolve the merge conflict before I review.

lemonboston:
updated with master, @dmfs please review

Issue #63, opened 2016-12-30
In order to improve our visibility we need to improve the current Powered by text


lemonboston:
@wilfredstegeman commented on [Fri Dec 30 2016](https://github.com/schedjoules/backlog/issues/98)

I suggest to move the current Powered by text to a sticky footer at the bottom:

- background color #000000
- text color #ffffff
- text in CAPITALS
- hyperlink the complete text with an option to change the link as we do with the feedback url

Screenshots

- Mockup SchedJoules
- Example HelpShift

![schedjoules-powered-by-footer](https://cloud.githubusercontent.com/assets/5968782/21562383/dc036ab8-ce78-11e6-8111-eba00805eb15.png)
![helpshift-powered-by](https://cloud.githubusercontent.com/assets/5968782/21562385/e14cf804-ce78-11e6-84ef-433c6025f5be.png)

---

@lemonboston commented on [Fri Dec 30 2016](https://github.com/schedjoules/backlog/issues/98#issuecomment-269790271)

I am not sure if this is agreed already, but I started to work on it.

dmfs:
Not a fan of this. Looks a bit desperate to me.

wilfredstegeman:
Can you elaborate on 'desparate'? As said before, we have to mention ourselves in partner apps, it's the best exposure and marketing of ourselves we can get. Moving the current text to the bottom and center is also an option, but that doesn't seem feasible. I'm open for other options, but the way it's currently shown has to change at least. 


		_____________________________

lemonboston:
@dmfs , @wilfredstegeman please check the apk in Slack with the black bottom bar, I put this on hold until solution is discussed.

dmfs:
I mean that it looks as if we badly need some attention. We should show enough self-confidence to not rub it in the users' faces all the time where this is coming from.

Our use case can not be compared to the Helpshift use case. As a regular user you only see the Helpshift screen when you need some support, i.e. something is wrong, doesn't work as expected or any other problem occurred. So the goal of an app developer should be that actually no user ever needs to see that Helpshift screen.
That being said, even in their use case I would refrain from that ugly bar.

In our case, the goal is that the user returns as often as possible. That's a fundamental difference. The black bar on the bottom almost reminds me on the typical ad banners. It looks desperate and cheap and I think that many users feel the same (especially since it's pretty uncommon on Android).

[...]

PR #64, opened 2017-01-02
PR for 63 - Powered by SchedJoules footer as fix black bottom bar

lemonboston:
#63 

Not ready for review, pushed only for showcase and to keep this state.

TODOs if continued:

- check if programmatic layout view decoration could be used to add the footer, to avoid the need to put it into every layout
- all micro-fragments of the details view have been updated, check if all looks okay
- make the url dynamic (can be changed from server without client change, as mentioned in the story)


lemonboston:
As this way has been discarded for a better solution, I am closing it and deleting the branch.

PR #71, opened 2017-01-05
Switch from Dumbledore to MicroFragments

dmfs:
Update package structure and move a couple of classes to the correct package
Implement WebviewMicroFragment to present websites in the context of the app.

@lemonboston please don't be scared by the number of modified files and the size of the diff. I've just moved a couple of classes to new packages. The actual code changes are quite small.


lemonboston (approved):
I notice a little glitch or lag when the feedback screen is sliding in, in the toolbar (the text 'Feedback').
Other than this, it looks good to me.

dmfs:
Fixed the glitch by delaying the 2nd transition until the first one finished.
Merging now.

Issue #76, opened 2017-01-09
Upgrade MicroFragments


dmfs:
upgrade to latest MicroFragments commit

PR #78, opened 2017-01-10
upgrade MicroFragments #76

dmfs:
fix imports
Note: we can't move the loaders to onCreate because the FragmentManager might call it even if the fragment is not visible at all.
try to save the webview state (not really related to the upgrade though)


lemonboston (approved):
Looks okay to me

PR #83, opened 2017-01-11
Improve "powered by" footer by adding a special FooterBehavior #63

dmfs:
Also load the entire footer from an image and add a grey background.


lemonboston:
Wow, nice, advanced solution with this custom CoordinatorLayout Behaviour and anchoring. I don't know much about them, so I can't properly review that part without spending more time, which is not necessary I suppose.

I tested with adding long event description to see the scrolling, it works fine.
The gap between the footer and Direction is a bit small I think, also the top gap between the bar and the vertical items looks a bit too big. I tried with these adjusments and looked better for me:
image

lemonboston (approved):

PR #84, opened 2017-01-12
update MicroFragments to fix #75

### LINKED ISSUE #75: Feedback menu option not working when cancelling an action at the chooser.
When clicking an action that results in a chooser and cancelling the launch, the feedback menu option no longer works.

dmfs:
When Android shows a chooser, the Fragment is not stopped but just paused. When you cancel the chooser it's simply resumed. That caused a problem in the MicroFragmentHost which didn't reset the `instanceStateSaved` flag in `onResume` but only in `onStart`, hence all incoming transitions were queued instead of executed.


lemonboston (approved):
tested the mentioned case and it works

Issue #68, opened 2017-01-05
In order to display parsed html properly SDK needs render HTML


lemonboston:
@rutgerg commented on [Tue Jan 03 2017](https://github.com/schedjoules/backlog/issues/106)

---

@rutgerg commented on [Tue Jan 03 2017](https://github.com/schedjoules/backlog/issues/106#issuecomment-270160346)

Can we make a html renderer to pick and choose which tags to parse and how? Eg convert all <hx> to <h5>?

---

@dmfs commented on [Tue Jan 03 2017](https://github.com/schedjoules/backlog/issues/106#issuecomment-270223018)

The easiest way is probably to use Android's `Html.fromHtml` method and process the resulting `Spannable` in order to remove/replace all styles we don't support.

@lemonboston any other ideas?

---

@lemonboston commented on [Wed Jan 04 2017](https://github.com/schedjoules/backlog/issues/106#issuecomment-270363928)

I am not sure, is it possible to remove styles from `Spannable`?

But I would like ask some questions first:

- Do we do this processing on the client side then? Is it only for easier solution because Android already has html renderer?
- I wasn't convinced myself yet if rendering the html instead of showing the plain text extract is better, but I haven't worked with htmls much. Could we maybe build some test versions of the app that points to an endpoint that uses Meetup events so that all 3 of us can actually check how they look like? (it can be done quickly)
Btw, is there maybe a Dropbox team account or anything else where we can quickly put up apks, screenshots, screen videos? Or should we make Slack channels for it? (not sure about it's storage limit)
I would also check on meetup.com, how a description is made and check examples to have a better picture.

---

@dmfs commented on [Wed Jan 04 2017](https://github.com/schedjoules/backlog/issues/106#issuecomment-270402760)

We're talking about two steps here:
- sanitize the HTML, i.e. remove/replace all unsupported tags
- render the sanitized HTML and apply custom styles

As I mentioned somewhere else, ideally the server already returns this sanitized HTML, so we don't need to do that on the client side.

However, we still might want to influence the way it's actually rendered by TextView. That also includes the font size of headlines.

---

@lemonboston commented on [Wed Jan 04 2017](https://github.com/schedjoules/backlog/issues/106#issuecomment-270417419)

@rutgerg Do you maybe have a server endpoint or request parameter that we can use to check the Meetup events?

---

@lemonboston commented on [Wed Jan 04 2017](https://github.com/schedjoules/backlog/issues/106#issuecomment-270422248)

[...]

PR #85, opened 2017-01-12
Parse event description as HTML. Initial implementation of #68

dmfs:


lemonboston (approved):
Looks okay to me.
Maybe we could change the link color from red to something else?
Also maybe in the next round the ImageGetter could be added to the Html call, I searched for GlideImageGetter and there are code samples for it (unfortunately it's not just a one liner call it seems).

dmfs:
I think links use the accent color by default. I need to check how to change that. But we certainly should.

We're not going to support images in the description. The server should strip `` tags.

dmfs:
I've updated the code to use the primary color for links instead.  @lemonboston please check it out.

lemonboston:
Checked, it is better with this color imo, too.

The deprecated getColor() method call can be replaced with `ContextCompat`:
`ContextCompat.getColor(getContext(), R.color.schedjoules_colorPrimary)`

PR #86, opened 2017-01-13
Add powered-by footer to event list, alternative to #63

dmfs:
@lemonboston please review. We're going to add this to the list for now, but we're certainly going to add some behavior later on.

### DIFF
--- a/eventdiscovery-sdk/src/main/res/layout/schedjoules_event_list.xml
+++ b/eventdiscovery-sdk/src/main/res/layout/schedjoules_event_list.xml
@@ -15,8 +15,10 @@
                 android:layout_width="match_parent"
                 android:layout_height="match_parent"
                 app:layoutManager="LinearLayoutManager"
+                android:layout_marginBottom="24dp"
                 tools:context=".framework.eventlist.EventListActivity"
                 tools:listitem="@layout/schedjoules_list_item_event"/>
 
+        <include layout="@layout/schedjoules_powered_by_footer"/>
     </FrameLayout>
 </layout>
--- a/eventdiscovery-sdk/src/main/res/layout/schedjoules_fragment_event_details.xml
+++ b/eventdiscovery-sdk/src/main/res/layout/schedjoules_fragment_event_details.xml
@@ -62,30 +62,19 @@
                 <View
                         android:id="@+id/schedjoules_footer_placeholder"
                         android:visibility="invisible"
-                        android:layout_height="20dp"
+                        android:layout_height="24dp"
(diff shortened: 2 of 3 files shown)


eventdiscovery-sdk/src/main/res/layout/schedjoules_event_list.xml
```diff
@@ -12,9 +12,27 @@
                 android:layout_width="match_parent"
                 android:layout_height="match_parent"
                 app:layoutManager="LinearLayoutManager"
+                android:layout_marginBottom="20dp"
```
lemonboston: I checked with layout inspector and this margin leaves a little bit too much space covered by the footer, can be changed to 24dp safely (I checked), to avoid incorrect proportions of the bottom items (no more events).

eventdiscovery-sdk/src/main/res/layout/schedjoules_event_list.xml
```diff
@@ -12,9 +12,27 @@
...
                 tools:context=".framework.eventlist.EventListActivity"
                 tools:listitem="@layout/schedjoules_list_item_event"/>
 
         <include layout="@layout/sticky_header_layout"/>
+
+
+        <com.schedjoules.eventdiscovery.framework.microfragments.eventdetails.fragments.views.SchedJoulesFooterView
```
lemonboston: I think the duplication should be resolved with the footer in the details view, either with layout of its own and `include` tag (extra tags can be added to `include`), or extracted style, but probably `include` is better, imo.

lemonboston (dismissed):

dmfs:
@lemonboston please review again. Note that I've rebased and squashed the commits, so you should remove your local branch first.
The problem was that the text size wasn't set to 0 which means the bar was slightly larger than the image+padding. Setting the textSize to 0 and increasing the padding fixed it.

lemonboston:
Why is `SchedJoulesFooterView` still a `TextView`? Couldn't it be simply an `ImageView`?

I think it would be useful to move the common layout definition of `<SchedJoulesFooterView....>` to a separate layout file and include it to remove the duplication. And use it like:
```
<include layout='schedjoules_powered_by_footer'
         app:layout_anchor="@id/schedjoules_footer_placeholder"
        app:layout_anchorGravity="bottom|center_horizontal"/>
```
So adding the special properties at the place of include.

dmfs:
If we use an `ImageView` we have to wrap it in a `FrameLayout` to get the padding and background color right. I agree that a `TextView` looks strange but it serves the purpose.
Good idea with the `include`. I'll do that.

Issue #21, opened 2016-12-07
Create manual or automated test cases for checking the various event list behaviors


lemonboston:
There are many different scenarios for event list behaviors which are hard to test with real data, and would also be difficult if not impossible to test them with unit test.
Making sort of manual test suites using certain dummy data setups and describing a checklist to look through can help us make sure the list behaves correctly.

Some notes on the implementation:
- DummyEvents (-> rename to Mock*) and its contents could be moved to demo app debug* classpath and used in a Mock api service, so it won't be in the SDK. Inner classes can be made top level.
- Improve on them to easily parameterize a setup for the mock data (eg: 3 pages down until end, error 4 times on page 2, number of events per page, hours between events, etc)
- Find a way to create and document the test-cases at a place in a Java class. Create and test them.

(*edit: debug classpath instead of test classpath)

lemonboston:
I mentioned demo app-debug classpath because I remembered it's not possible for library projects to use debug and release versions, but I am not sure now, so that's worth checking as well.

lemonboston:
Just a note here that I tried cloud testing with my personal account but I get these error (for multiple days now, so it never run successfully):
```
Testing started at 1:30 PM ...
Creating Cloud Storage Bucket as-build_eventdiscovery-sdk_1483014604752_orxa ...
Uploading app APK ...
Uploading test APK ...
Submitting tests to Firebase Test Lab ...
Validating APKs ...

You can also view test results, along with other runs against this app, on the web:
 

Nexus 5, LG, Virtual | Android 6.0.x, API Level 23 (Marshmallow) | English | Landscape
	Infrastructure Failure: Internal error during validation
Test running failed: No test results
Finish

Pricing information: https://firebase.google.com/pricing 
```
I couldn't find any useful information on this error.

Issue #20, opened 2016-12-07
Create unit test for event list items updates (merge, add, remove, ..)


lemonboston:
Unit test for current `EventListItemsImpl`.

lemonboston:
Pending on branch `20-unit-test-for-listitems-merge` until https://github.com/schedjoules/android-event-discovery-sdk/pull/32 is merged.

PR #103, opened 2017-01-25
PR for 70: Update FlexibleAdapter

lemonboston:
Updated FlexibleAdapter dependency to latest 5.0.0-rc1 and also excluded support libraries, so the effective support lib version of our SDK changed from 25.0.1 to 25.0.0 because of this. (I think multilinecollapsingtoolbar pulls in 25.0.0).

Also changed the usage a bit.

@dmfs please review
Please leave the story open after this has been merged.
(There will be another bigger refactor PR coming to update usage according to FlexibleAdapter wiki, namely to not add the headers to the list. I also changed multiple related classes there, it's on a local branch now, I will need the PRs merged from stories #20 and #21 before I can continue with that, in order to test the list behaviors.)

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/flexibleadapter/FlexibleAdapterNotifier.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/flexibleadapter/FlexibleAdapterNotifier.java
@@ -42,16 +42,18 @@ public final class FlexibleAdapterNotifier implements ThirdPartyAdapterNotifier<
     @Override
     public void notifyInitialItemsAdded(List initialItems)
     {
-        //noinspection unchecked
-        mFlexibleAdapter.addItems(0, initialItems);
+        notifyNewItemsAdded(initialItems, 0);
     }
 
 
     @Override
     public void notifyNewItemsAdded(List newItems, int positionStart)
     {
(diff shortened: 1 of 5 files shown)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/flexibleadapter/FlexibleAdapterNotifier.java
```diff
@@ -42,14 +42,17 @@ public FlexibleAdapterNotifier(FlexibleAdapter flexibleAdapter)
...
-        //noinspection unchecked
-        mFlexibleAdapter.addItems(0, initialItems);
+        notifyNewItemsAdded(initialItems, 0);
     }
 
 
     @Override
     public void notifyNewItemsAdded(List newItems, int positionStart)
     {
+        if (newItems == null || newItems.isEmpty())
```
dmfs: Could newItems possibly be null here? I.e. do you ever willingly pass `null` to this method directly or indirectly?
If not, I'd rather see this fail here, because there is clearly something wrong.

lemonboston: Right, it was the constructor that was called with null before, but this method should not be called  with null. Removed the null check.

PR #104, opened 2017-01-27
PR for 93: Location search 'X' button

lemonboston:
@dmfs please review


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/ClearableEditText.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    public ClearableEditText(Context context, AttributeSet attrs, int defStyleAttr, int defStyleRes)
+    {
+        super(context, attrs, defStyleAttr, defStyleRes);
+        init();
+    }
+
+
+    private void init()
```
dmfs: Isn't `onFinishInflate` a more appropriate place for this kind of initialization?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/ClearableEditText.java
```diff
@@ -0,0 +1,97 @@
...
+            }
+        });
+
+        // http://stackoverflow.com/a/29214019/4247460
+        setOnTouchListener(new View.OnTouchListener()
+        {
+            @Override
+            public boolean onTouch(View v, MotionEvent event)
+            {
+                if (event.getAction() == MotionEvent.ACTION_UP)
```
dmfs: IMO this approach is a bit too simple, because it will wipe the field even if you don't tap on the "X" but swipe onto it before lifting your finger. You should at least test if the finger hit the area in the previous `ACTION_DOWN` event.

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/ClearableEditText.java
lemonboston: Right, probably better, changed it.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/ClearableEditText.java
lemonboston: Yes, thanks for noticing, changed it to use `GestureDetector` to detect the tap.

lemonboston:
@dmfs I've made the changes, please check again.

dmfs (approved):

Issue #101, opened 2017-01-20
In order to have a nicer location marker on the location screen, as a user I want the current menu icon changed


lemonboston:
It can be a drop down arrow next to the location name or at the menu icon location or also location icon (pin) there instead of the current icon.

wilfredstegeman:
I suggest to use a drop down arrow (open or closed) next to the location. This is also used by Meetup and Eventbrite, feels very natural. In contrast to for instance 'All events in city' where they've disconnected the two with a location picked icon in the right top corner. 


		


_____________________________

lemonboston:
Started working on adding the drop-down arrow next to the title.

Work items checklist (updated continuously):

- [x] Check what are the options for implementation -> 1. Adding unicode character 2. Using ImageSpan and icon 3. Maybe some special fonttype/typeface

- [x] Try unicode character -> none of them looked good exactly and they might change if device use different font, also it strangely took 1-2 sec to load the screen after this change (maybe Android initializes a bigger character set?) -> solution dropped

- [x] Try with ImageSpan -> works

- [x] Create proper class to handle text and icon -> TextWithIcon created (custom spannable string design may be improved later if needed)

- [x] Remove location icon from menu

- [x] Bug noticed: slight vertical misalignment of the title -> cause: the icon is a bit bigger than the text -> options to fix: 1. fix alignment to center (-> didn't work) 2. resize the icon asset 3. resize the image in code 4. try to use vector asset instead of the png -> vector asset worked

- [x] Refine drop-down vector icon geometry

- [x] Change Toolbar to not fire onClick for the whole length, only for the title -> need to add a TextView inside the Toolbar, can't change this on Toolbar itself

- [x] Update left margin/padding adjustment logic for the case when there is no UP arrow, for the new TextView title -> after a first hack, found contentInset property of Toolbar which is exactly for this

- [x] Handle long titles (single line with dots) -> used android:ellipsize="middle", because "end" causes drop down to not shown fully, also it might make sense to use middle, we don't know which part of a long venue name is important, start or end

- [x] Double-check on earlier Android versions

PR #107, opened 2017-01-30
PR for 101: Drop-down arrow next to location name

lemonboston:
@dmfs please review.

Work items in the story: https://github.com/schedjoules/android-event-discovery-sdk/issues/101#issuecomment-276049981

The `TextWithIcon` class could be improved, didn't want to spend time with it now, but let me know if we should.
(I started with implmeneting `CharSequence`, but `Toolbar.setTitle(CharSequenc)` behaved differently if the parameter was `CharSequecne` or `SpannableString`, for `CharSequence` it simply called `toString()`, so the icon was not added. So there is some instanceof somewhere, so we would probably need to implement `CharSequence` and `Span` or `Spannable` together. An abstract class could be created with a factory in constructor to have lazy evaluation.)

Another possible generalization could a `ToolbarWithClickableTitle`, to extract the Toolbar and it's inner TextView, but it didn't seem that important either. Let me know I should do that.


eventdiscovery-sdk/src/main/res/values/dimens.xml
```diff
@@ -13,4 +13,5 @@
     <dimen name="schedjoules_list_item_default_height">96dp</dimen>
     <dimen name="schedjoules_event_list_progress_bar">36dp</dimen>
     <dimen name="schedjoules_event_detail_item_icon">28dp</dimen>
+    <dimen name="schedjoules_toolbar_title">20sp</dimen>
```
dmfs: Please give this a more meaningful name. `schedjoules_toolbar_title` doesn't say anything. Is this the height, width, padding, margin, font size?
I suggest the following name scheme for dimensions: `schedjoules_<context>_<element>_<attribute>`, with each block written in camel-case. In this case this would result in `schedjoules_toolbar_title_fontSize` or (depending what you define as the "context": `schedjoules_eventList_toolbarTitle_fontSize`)
`schedjoules_event_list_progress_bar` would become `schedjoules_eventList_progressBar_width`.
What do you think?
Btw is it good practice to define the Toolbar title in `sp`? Isn't the Toolbar fixed to `64dp`? In theory, setting a title in `sp` could result in a title that is larger than the toolbar.

dmfs (changes requested):

eventdiscovery-sdk/src/main/res/values/dimens.xml
lemonboston: You are right, that wasn't a good enough name, changed it and also changed `sp` to `dp`, I looked it up incorrectly earlier it seems, because I just wanted to match it with the theme (it was needed for a bug about smaller text size when screen rotated). But I clicked through now and the android theme has:
```
    <dimen name="text_size_title_material_toolbar">20dp</dimen>
```

Regarding the naming format, I agree and like the way to combine underscore and camel-case, but I would suggest we create a separate story for it to update every resource. First, define it for all resource types and then implement when there is not much parallel work in progress task, to avoid the conflicts.
I actually would like the first letters capital as well, it looked strange first when I saw that, but after I got used to it looked better actually, so something like this:
`SchedJoules_EventList_ProgressBar`
but this context first may not be best for all resource type I think, for example the order above for the text size in android theme might make sense also as well. So I think we should just discuss and defined the formats in a backlog story first. Okay to create it?

(To not mix formats, I renamed this dimen to `schedjoules_toolbar_title_font_size` now.)

lemonboston:
@dmfs I've made the update, please review again

dmfs (approved):

dmfs:
Before we create a story to rename the resources we should have a story to create a wiki page to document our naming scheme.

PR #109, opened 2017-01-30
PR for 108: Request uri logger

lemonboston:
@dmfs please review.
(Feel free to change things yourself, too.)

### DIFF
--- a/eventdiscovery-demo/src/main/java/com/schedjoules/eventdiscovery/demo/ApiService.java
+++ b/eventdiscovery-demo/src/main/java/com/schedjoules/eventdiscovery/demo/ApiService.java
@@ -8,9 +8,11 @@ import com.schedjoules.client.Api;
 import com.schedjoules.client.SchedJoulesApi;
 import com.schedjoules.client.SchedJoulesApiClient;
 import com.schedjoules.client.utils.StringAccessToken;
-import com.schedjoules.eventdiscovery.service.AbstractApiService;
+import com.schedjoules.eventdiscovery.framework.http.RequestUriLogging;
 import com.schedjoules.eventdiscovery.framework.utils.SharedPrefsUserIdentifier;
+import com.schedjoules.eventdiscovery.service.AbstractApiService;
 
+import org.dmfs.httpessentials.client.HttpRequestExecutor;
 import org.dmfs.httpessentials.executors.retrying.Retrying;
 import org.dmfs.httpessentials.executors.retrying.policies.DefaultRetryPolicy;
 import org.dmfs.httpessentials.executors.useragent.Branded;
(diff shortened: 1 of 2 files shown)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/service/RequestUriLogger.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+import java.net.URI;
+
+
+/**
+ * {@link HttpRequestExecutor} decorator that logs request uris to logcat.
+ *
+ * @author Gabor Keszthelyi
+ */
+public class RequestUriLogger implements HttpRequestExecutor
```
dmfs: Please rename to `RequestUriLogging`. This is a decorator and it should follow our decorator naming scheme, i.e. the decorator name describes how it modifies the original behavior. It's not an "-er" type of class.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/service/RequestUriLogger.java
```diff
@@ -0,0 +1,63 @@
...
+        mTag = tag;
+    }
+
+
+    @Override
+    public <T> T execute(URI uri, HttpRequest<T> request) throws IOException, ProtocolError, ProtocolException, RedirectionException, UnexpectedStatusException
+    {
+        if (mShouldLog)
+        {
+            Log.d(mTag, "SJ-ED request: " + uri);
```
dmfs: You can easily log the request method like so:
```java
Log.d(mTag, String.format("%s %s", request.method().verb(), uri));
```
That would result in something like
```
GET https://api.schedjoules.com/events/?u=xxxxxxx...
```

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/service/RequestUriLogger.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private final HttpRequestExecutor mDelegate;
+    private final boolean mShouldLog;
+    private final String mTag;
+
+
+    public RequestUriLogger(HttpRequestExecutor delegate, boolean shouldLog, String tag)
+    {
+        mDelegate = delegate;
+        mShouldLog = shouldLog;
```
dmfs: I don't really like variables that contain "should" because "should" implies that you don't "have to". Please use a more direct and clear language like `enableLog` or  `writeToLog`.

dmfs (changes requested):
Also, please move the logging decorator to a more appropriate package. Even though it may be used by partners, it's mostly for internal use and should go to a `framework.http` package.

lemonboston:
@dmfs Thanks, I've made the changes, please check again.

dmfs (approved):

PR #111, opened 2017-02-01
Fix displaced footer in details view. Fixes #110

### LINKED ISSUE #110: Footer in details screen displaced

dmfs:
Commit 53f9be195f1c moved the footer into a separate layout without merge tags. Apparently that didn't play well with the footer-behavior on the details screen.


lemonboston (approved):
I've checked, this works, but if I simply added `android:layout_gravity="top|end|right"` to the `include` in the details layout on master, it also seemed to work correctly for me.

Issue #114, opened 2017-02-02
Add demo app flavor that always enables request logs.


dmfs:
In order to publish an internal demo app with request logs, as an app publisher I want a flavor that always writes request logs, even in the release build.

Debug builds can't be published on Google Play, hence we have to enable request logs in the production build. But we should create a separate flavor for that to make sure someone who copies the ApiService code doesn't have request logs enabled in his production code.

dmfs:
It turned out there is an easier way by adding a local setting.

PR #118, opened 2017-02-03
Add a BuildConfig field that determines whether to log requests or not. Implements #114

dmfs:
By default request logging is enabled for debug builds and disabled for release builds.
To override this behavior add the following line to your `local.properties` file:

request.log.enabled=true


lemonboston (approved):
Looks good for me.

Issue #113, opened 2017-02-02
Start services non-sticky


dmfs:
In order to reduce resource usage and get better session information, as a developer I want all services to be started non-sticky.

PR #119, opened 2017-02-03
Start Action service and Insights service non-sticky. If they are kil…

dmfs:
…led they'll be restarted automatically when we need them. Implements #113

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/services/BasicInsightsService.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/services/BasicInsightsService.java
@@ -91,6 +91,14 @@ public final class BasicInsightsService extends Service implements InsightsServi
 
 
     @Override
+    public int onStartCommand(Intent intent, int flags, int startId)
+    {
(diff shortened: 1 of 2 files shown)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/services/BasicInsightsService.java
```diff
@@ -91,6 +91,15 @@ public void run()
 
 
     @Override
+    public int onStartCommand(Intent intent, int flags, int startId)
+    {
+        super.onStartCommand(intent, flags, startId);
```
lemonboston: I think the calls to `super` are probably not needed.
(The onStartCommand() call in super is only for pre 2.0 compatibility.)

lemonboston: Otherwise, looks good for me.

dmfs: Good point. I'll remove it.

Issue #112, opened 2017-02-02
Enable hardware acceleration by default.


dmfs:
In order to get a smooth experience, as a user I want the SDK to use hardware acceleration whenever possible.

Some partner apps might have disabled HW acceleration by default. We'll enable it explicitly for our activities in the AndroidManifest.

lemonboston:
Duplicate of https://github.com/schedjoules/android-event-discovery-sdk/issues/92, I suppose that one can be closed.

dmfs:
Oh, yes of course. I'll close the other one.

PR #120, opened 2017-02-03
Make sure our apps are hardware accelerated by setting the flag for e…

dmfs:
…ach visible activity.

This will override any default setting of the hosting application.
Implements #112


lemonboston (approved):
Looks good.

PR #122, opened 2017-02-03
Determine demo app version from Git. #121

dmfs:
The app version name will match the tag.
The app version code is derived from the number of commits on the master branch. Not sure if that causes problems when trying to build and publish a feature branch though. In that case we might have to intervene manually.

### DIFF
--- a/eventdiscovery-demo/build.gradle
+++ b/eventdiscovery-demo/build.gradle
@@ -9,6 +9,24 @@ if (googlePlacesApiKey == null) {
     println("WARNING! Missing Google Places API key in 'local.properties' with key 'google.places.api.key'. Location search won't work.")
 }
 
+def gitVersion = { ->
+    def stdout = new ByteArrayOutputStream()
+    exec {
+        commandLine 'git', 'describe', '--tags'
+        standardOutput = stdout
+    }
+    return stdout.toString().trim()
+}
+
+def gitCommitNo = { ->
+    def stdout = new ByteArrayOutputStream()
+    exec {
+        commandLine 'git', 'rev-list', '--count', 'master'
+        standardOutput = stdout
+    }
+    return Integer.parseInt(stdout.toString().trim())
+}
+
 android {
     compileSdkVersion 23
     buildToolsVersion "25.0.0"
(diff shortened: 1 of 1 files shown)


dmfs:
I think the same mechanism would come in handy for our regular app too.

lemonboston (approved):
This is great! I like how the version number can also be automatized with this git commit count.

I think it would be worth extracting to a separate `.gradle` file and import with `apply from: '<>.gradle'`.

Regarding releasing from feature branch, maybe we could do something like 'spreading out' the version number, so multiplying by 10 for example, so instead of 176, it wold be 1760 and then if we are on a feature branch, it could add the count of commits on the feature branch to that.
Or somehow throw an exception when we would build a release on feature branch, from the git function.
But none of them worth it probably to spend time with, so I think it's fine as you wrote that we do this manually when releasing from feature branch.

dmfs:
@lemonboston please re-review.

eventdiscovery-demo/build.gradle
```diff
@@ -9,15 +9,33 @@ if (googlePlacesApiKey == null) {
...
     compileSdkVersion 23
     buildToolsVersion "25.0.0"
     defaultConfig {
         applicationId "com.schedjoules.eventdiscovery.demo"
         minSdkVersion 14
         targetSdkVersion 23
-        versionCode 1
-        versionName "0.1"
+        versionCode gitCommitNo()
+        versionName gitVersion() * 10 // spread the version number so we can insert versions if necessary
```
lemonboston: Missed one line probably.

I was curios to see what is _string_ x 10, this :) :
```
android:versionCode = 174
android:versionName = 1.0-6-g37e34481.0-6-g37e34481.0-6-g37e34481.0-6-g37e34481.0-6-g37e34481.0-6-g37e34481.0-6-g37e34481.0-6-g37e34481.0-6-g37e34481.0-6-g37e3448
```

lemonboston (changes requested):

eventdiscovery-demo/build.gradle
dmfs: hehe :-D
ahem, this was just a test to see if you pay attention to details ;-D

dmfs:
@lemonboston please re-re-review

lemonboston (approved):

lemonboston:
Looks good.

Issue #70, opened 2017-01-05
Stuck top day header


lemonboston:
What I was doing:
List screen with Amsterdam, open location selector, select Rotterdam, coming back to List screen, it updated to this:

![double-tomorrow](https://cloud.githubusercontent.com/assets/1077972/21694252/8fb44ac2-d384-11e6-961d-7a376591492d.png)

dmfs:
Had a similar issue today:
![screenshot_2017-01-09-17-26-02](https://cloud.githubusercontent.com/assets/1614987/21787153/dee4b26c-d6c7-11e6-90c6-7046265e5f00.png)

lemonboston:
Some foundings so far:

After about 15 minutes of trying I could reproduce the bug, the screeshot:
![double-header](https://cloud.githubusercontent.com/assets/1077972/21933042/f2bc5ec8-d9a3-11e6-8d57-10f29fb463c3.png)

I could debug the code and also inspect the view hierarchy and that stuck header element is not there in the adapter's list (not in FlexibleAdapter's internal list and not in our own list), so the lists are correct, but it is the first child of the RecyclerView in the view hierarchy.
It stays there even after selecting another location again.

Here is the comparison of the attributes of the FrameLayout showing this stuck day header and another FrameLayout that is a correct day header in the list:
![compare2](https://cloud.githubusercontent.com/assets/1077972/21933350/5c5cce2a-d9a5-11e6-8629-a4dcb2fda45a.png)

What may be interesting is that FORCE_LAYOUT flag in the stuck header.
In `ViewGroup`, there is this usage:
```
    /**
     * Prevents the specified child to be laid out during the next layout pass.
     *
     * @param child the child on which to perform the cleanup
     */
    protected void cleanupLayoutState(View child) {
        child.mPrivateFlags &= ~View.PFLAG_FORCE_LAYOUT;
    }
```
So I suppose it might suggest that something went wrong with the layout() cycles.

A screen video of the behaviour of this stuck header:
https://drive.google.com/open?id=0Bzp0PP6UbK4lS3BTN1ZQN3BKSkk
The view hierarchy file, can be opened in Android Studio:
https://drive.google.com/open?id=0Bzp0PP6UbK4lSUgzekRIV2lrN3M

lemonboston:
The FlexibleAdapter is currently cleared by calling this:
```
mFlexibleAdapter.removeRange(0, totalSize);
```
[...]

PR #124, opened 2017-02-07
PR for 70: Refactor all FlexibleAdapter usage

lemonboston:
Relates to #70.

@dmfs Please review.
The goal of this refactor was to match our usage of FlexibleAdapter to their wiki, namely to not add `IHeader` items explicitly to the list. But during working it out several other changes were needed or were benefitial:
- changing how we adapt our `ListItem`s to FlexibleAdapter types
- removing the separation of notiying the adapter and keeping the list of data, because FlexibleAdapter has its own list anyway, and it made things more straightforward


dmfs:
In general it works well but I noticed a glitch when jumping to a date in future at a specific time. If there are other events before that time (but on the same day) on the previous page the list will jump to the first event of that day when it should stick with the event that was on top before.

I'll post a screen recording in Slack.

lemonboston:
Yes, I knew about that.. it doesn't happen if you just scrolled away already, but otherwise it does.
This is how it works with the `addItemToSection()` call of FlexibleAdapter.
I tried to fix this but I couldn't in about an hour, so currently the only way I see to solve this is to revert back to adding the Headers as well to the list, so basically same as before. This refactor would still carry some value, but the basic thing would be reverted, which I don't know if it's important or not, cause it worked that way as well.

Updated my previous question on FlexibleAdapter github, maybe we get some tip from him:
https://github.com/davideas/FlexibleAdapter/issues/277
I suggest we put this on hold for now.

dmfs:
Ok, on hold it is.

lemonboston:
I found a fix!
I needed to remove the existing header from `FlexbleAdapter` when merging to top and it fixed it.
Hallelujah :)
@dmfs Please review.

lemonboston:
Asked question about flashing header here:
https://github.com/davideas/FlexibleAdapter/issues/277#issuecomment-278937049

lemonboston:
I possibly found a (not very nice, but working fix) for the header flash when items added on top without merging. I plan to add it to this PR today.

lemonboston:
The supposed fixed I found for the flashing header didn't work consistently, so I didn't apply it.

And I now checked against `master` version: there the flash doesn't happen when items loaded on top, at least this time - because the whole thing seems inconsistent sometimes, not sure if it is deterministic.
@dmfs Could you please check this yourself as well? Comparing this branch and master for flashing. 
Because I suppose we can't merge this if this version makes more 'flashes' than `master`.

lemonboston:
I've found a hacky way to fix the flash for the top header for the case when there is no merging needed (meaning no common header in the current list and the incoming items). It works on the demo app, although on the manual UI test it still flashes.

I've just asked once more about it on the FlexibleAdapter github issue, so I suggest we wait if there is a reply that might help find a better solution. If there isn't, then I think we could merge this after a review.
(I've rebased to `master`, so the branch is kept in line with `master`)

lemonboston:
I've just applied another hacky improvement on the top items merging code based on the suggestion from FlexibleAdapter author.
So in spite of the hacky logic for the item mergings in `FlexibleAdapterEventListItems#mergeNewItemsTop` I consider this now finished.

@dmfs Please review.

dmfs:
I'm still having a minor glitch: When you scroll up quickly so you see the loading indicator for the previous page, the header is not shown after the items have been loaded. It only becomes visible while you scroll further up.
However, it's a minor issue and possibly a bug in FlexibleAdapter anyway. I'm merging this now.

PR #125, opened 2017-02-08
PR for 94: Location search list modularization

lemonboston:
#94

@dmfs It's ready for a first review / input.
Many things can change, especially names, I'll add javadocs, remove dummies, review everything again later.

One important question will be to use retained fragment or not, I suppose we should discuss it, for me it's not obvious what is the better way.
It affects the `SearchModule` interface.

One element that become key now and I want to mention is this concept of `ListChange`:
```
public interface ListChange<T>
{
    void apply(List<T> oldItems, RecyclerView.Adapter adapter);
}
```
So in order to make a change on a `RecyclerView`, we need to have access to the current (modifiable) list to update it and the adapter for the notification methods. Every kind of change can be encapsulate in a `ListChange` like this and can be created declaratively.
There is a component that holds the list and the adapter which can receive these `ListChange`s and apply them. (see `BasicChangeableListItems`)
I think it's a very useful concept, I could for example refactor the event list updates with this as well, even with `FlexibleAdapter`.

But I couldn't use this exactly for the modules directly because they should only know about their section of the list, so they cannot call adapter notifications, only the component that 'knows' the whole list can do that, especially with using `DiffUtils` now. So I introduced this poorly named thing:
```
public interface NonNotifyingListChange<T extends ListItem>
{
      void apply(List<T> oldItems);
}
```
So the modules work with this and there is an adapter that applies it to the whole list (items have section numbers in the whole list): `ListSectionChange`. It can surely be optimized more if needed, I just wanted to make it work at this stage.


dmfs:
Very nice.

I'd prefer if we go for a non-retained fragment. I don't think that serialization is that much of an issue. In general we're only talking about a few items that needs to be retained (also in most cases the items are not actually parcelled during the config change).

As far as it concern running background tasks during a configuration change, I'm not concerned about restarting these after the config change. In general they shouldn't run for more than a few seconds. If we have tasks that might have to transfer large amounts of data or take long for other reasons we'll outsource them to a service or content provider that can cache the result.

Consider to rename `ListChange` to `ChangeSet` instead, because that's what it represents. A change of the list is the result of applying a change set.

lemonboston:
@dmfs Please review again.
Even if you don't request change, please don't merge yet, Dummy modules are still in, I will remove them before merge.

Some notes:

Regarding the name `Change` or `ChangeSet`, I had thought about it before and for me `ChangeSet` means just a plural version of `Change`. "set of changes" as on wikipedia for the vcs meaning. So for me it means multiple changes and these changes we use are usually just simple ones, logically one change, like `ReplaceAll` and `ClearAll` (btw could be just `Replace` and `Clear`), or there may be `AddToTop` or `AddToBottom`, things like that. They can be combined of course, but usually they will be just one simple change. So that's why I stayed with `Change`.
_A change of the list is the result of applying a change set._ I am not sure I get this, for me either a `Change` or a `ChangeSet` represents a change for the list, created declaratively, that can applied at a later time to the list.
But it really doesn't matter much of course, so let me know if you prefer the names `ListChangeSet` and `NonNotifyingListChangeSet`, I will update them.

There is a new `SearchModule` factory now called `SelectableSearchModuleFactory` which takes an extra parameter of `ListItemSelectionAction`. I am not totally sure about this, so I am curios what you think. 
So this is intended to be used when all items in a search list will eventually react the same way to a click/selection. This may not be true for all search lists, but for location picker it is.
So this action is passed in, so the module can execute it when it wants to.
Would be of course nicer if this is handled outside of `SearchModule` but as `PlacesSuggestionModule` needs that extra fetch to get the eventual `GeoPlace`, it is now passed in to the `SearchModule`. The action now captures the `Activity`, so it has to be nulled out in `shutDown()`. This is where it is not very nice, but if there would be a callback, that would have to be nulled out, too. (Should we use WeekReference maybe?)
(Not sure about the following, but the benefit of this `ListItemSelectionAction` might be that it could be decorated/adapted or passed along. For example if the `ListItem` doesn't own the same data type but handles `ListItemSelectionAction`, it can receive one that wraps the one passed to the module and convert/adapt between the data types. The whole thing might connect later to generalizing `ListItem` usage more with `SmartView` and this `ListItemSelectionAction`. But this not something I worked out or see clearly already.)

lemonboston:
One more note: you've probably seen already that the list update animation is sometimes not very nice, it's when there are two subsequent changes right after each other, because the Recent Places reacts quicker than the Suggestions, so the update after a query come in 2 steps.
I am not sure this is something we ever want to do something with, like scheduling/synchronizing updates, because it's also seems okay in general that modules react independently, because one may take long and the other is immediate. So there could be some short cycle delay that merges all updates coming in during that time, or maybe better, delaying a change if it comes too quick after the previous one, but probably none of this is very easy to implement.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
```diff
@@ -0,0 +1,202 @@
...
+ * {@link SearchModule} for showing the Google Place API places suggestions.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class PlaceSuggestionModule implements SearchModule, PlaceSuggestionItem.OnClickListener
+{
+    public static SelectableSearchModuleFactory<GeoPlace> FACTORY = new SelectableSearchModuleFactory<GeoPlace>()
+    {
+        @Override
+        public SearchModule create(Activity activity, NonNotifyingChangeableList changeableList, ListItemSelectionAction<GeoPlace> itemSelectionAction)
```
dmfs: Where it the generic type of NonNotifyingChangeableList?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
```diff
@@ -0,0 +1,202 @@
...
+             * See https://developers.google.com/android/reference/com/google/android/gms/common/api/GoogleApiClient.Builder.html#enableAutoManage(android.support.v4.app.FragmentActivity, com.google.android.gms.common.api.GoogleApiClient.OnConnectionFailedListener)
+             */
+
+            googleApiClient.connect();
+            return new PlaceSuggestionModule(changeableList, itemSelectionAction, googleApiClient);
+        }
+
+    };
+
+    private static final String TAG = PlaceSuggestionModule.class.getSimpleName();
```
dmfs: Please use constant strings for TAGs. When using Proguard the class name might be obfuscated.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
```diff
@@ -0,0 +1,202 @@
...
+
+    private final NonNotifyingChangeableList<ListItem> mChangeableList;
+
+    private ListItemSelectionAction<GeoPlace> mSelectionAction;
+
+    private GoogleApiClient mGoogleApiClient;
```
dmfs: Please make all fields `final`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/searchlist/SelectableSearchModuleFactory.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param activity
+     *         can be used to initiate connections/resources that the {@link SearchModule} may need
+     * @param changeableList
+     *         the {@link SearchModule} can use this to update its section in the list.
+     * @param itemSelectionAction
+     *         should be executed by the {@link SearchModule} when the user selects an item
+     *
+     * @return new instance of {@link SearchModule}
+     */
+    SearchModule create(Activity activity, NonNotifyingChangeableList changeableList, ListItemSelectionAction<T> itemSelectionAction);
```
dmfs: Generic type of NonNotifyingChangeableList?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/searchlist/SearchModulesFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+
+/**
+ * Factory for creating {@link SearchModule}s for a search list.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class SearchModulesFactory implements Factory<List<SearchModule>>
```
dmfs: What's the difference to `SelectableSearchModulesFactory`? Why do we need both of them? In particular, what would be a use case for a search module without selectable items?

eventdiscovery-sdk/src/main/res/layout/schedjoules_list_item_dummy_header.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        android:paddingLeft="@dimen/schedjoules_list_item_horizontal_margin"
+        android:paddingRight="@dimen/schedjoules_list_item_horizontal_margin"
+        android:background="#d9d9d9"
```
dmfs: For consistency, please use the same header color as in the events list. We'll rework the design later on.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+
+    @Override
+    public void shutDown()
+    {
+        mGoogleApiClient.disconnect();
+
+        // TODO think about this, is callback better? That would have to be nulled out as well.
+        mSelectionAction = null;
```
dmfs: I don't think this needs to be `null`ed. If you have concerns about leaking a context use a `WeakReference`.

Btw. you can't use a `WeakReference` like so
```java
if (weakReference.get() != null)
{
    weakReference.get().doSomething(...);
}
```
There is no guarantee that the reference still exists between these two calls. In theory the garbage collector could collect the referenced object right between these two calls. So the second call could result in a NPE. You need to store the reference in a local variable:
```java
SomeType ref = weakReference.get();
if (ref != null)
{
    ref.doSomething(...);
}
```

Just mentioning this because I've just seen code like above in some of the async tasks.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+
+    private final class PlaceSuggestionQueryTaskClient implements PlaceSuggestionQueryTask.Client
+    {
+
+        @Override
+        public boolean shouldDiscard(String query)
+        {
+            return !Objects.equals(query, mLastQuery);
```
dmfs: IMO this shouldn't be handled here. In particular `mLastQuery` looks like an inappropriate piece of mutable state.

As suggested on slack I'd prefer a different control flow. Which is, a `SearchResult` is introduced that holds the actual results as well as the query that it was launched for. The `SearchResult` is then returned to the owner of the module. When applying the search result, the query will be checked against the current query.

To cancel any pending tasks you should take control over the execution queue:

```java
mJobQueue = new LinkedBlockingQueue<Runnable>();
mExecutorService = new ThreadPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS, mJobQueue)
```

Whenever a new query arrives you just wipe the queue to cancel pending requests.
```java
mJobQueue.clear();
```

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
```diff
@@ -0,0 +1,202 @@
...
+
+
+    @Override
+    public void shutDown()
+    {
+        mGoogleApiClient.disconnect();
+
+        // TODO think about this, is callback better? That would have to be nulled out as well.
+        mSelectionAction = null;
+    }
```
dmfs: don't forget to shut the executor down to cancel pending requests.
```java
  mExecutorService.shutdown();
```

dmfs (changes requested):
You're right the animations are not very pretty when many changes happen quickly. Is there any way to delay an update until the animation of the previous update has been completed?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
lemonboston: Added it.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
lemonboston: Thanks, good to know, updated it.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
lemonboston: Done

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
lemonboston: Ok, I don't bother with nulling out the action (without using the modules in a retained fragment it cannot actually cause context leak.)

Thanks for the note on `WeakReference` usage, updated the `AsycnTasks`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
lemonboston: Good catch, thanks, updated.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/searchlist/SelectableSearchModuleFactory.java
lemonboston: Added.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/searchlist/SearchModulesFactory.java
lemonboston: The example for a search list with elements that react differently for selection would be the Google search on the phone, it can show both apps and web search results as well.
But you are right, it's probably not worth keeping, maintaining unnecessarily, we can create it if we need later. So I removed it and removed `Selectable..` prefix from that one, so this is the default usage now.

eventdiscovery-sdk/src/main/res/layout/schedjoules_list_item_dummy_header.xml
lemonboston: Ok, unified the two headers.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/PlaceSuggestionModule.java
lemonboston: The problem with the `SearchResult` approach was that the module may need to update the list not only reacting to queries, but for example removing and item from Recent Places list by tapping an 'X' button.
Continue this part on Slack..

lemonboston:
I made the changes to use `ResultUpdate` and a jobqueu in `PlacesSuggestionModule`.

I've also created a delaying decorator for the `ResultUpdate` listener. It has two delays:
- one is to wait a little for others to join (10 ms currently) - it was needed because otherwise the second instant reacting module would have to wait for full animation to finish even though they arrive within a millisecond
- the other is how much to wait after the previously fired update to fire the new ones, so to wait for animation to finish (400 ms now)

There is no aggregating mechanism currently, just simply firing the changes independently one after each other. It seems to work, although it would of course be more optimal to aggregate them. That requires some refactor, because applying an individual section change to the whole list is hidden now under one class.

The new `SearchListItems` may also have a bit too much responsibility, interface. I am not sure, because it's also nice that both the list item and the current query is kept there, so the states are there.
@dmfs Please have a look and let me know if you have any suggestions for changes, I will get back to it tomorrow anyway.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/LocationSelectionFragment.java
```diff
@@ -71,25 +77,34 @@ public static Fragment newInstance()
...
-         * Either enable automanage or add 'manual' error handling with addOnConnectionFailedListener().
-         *
-         * See https://developers.google.com/android/reference/com/google/android/gms/common/api/GoogleApiClient.Builder.html#enableAutoManage(android.support.v4.app.FragmentActivity, com.google.android.gms.common.api.GoogleApiClient.OnConnectionFailedListener)
-         */
-        mGoogleApiClient.connect();
-
-        mSuggestionItems = new BasicChangeableListItems();
-        mPlaceListController = new PlaceListControllerImpl(mGoogleApiClient, this, mSuggestionItems);
+
+        mSearchListItems = new UpdateDelaying(new BasicSearchListItems(), 10, 400);
```
dmfs: It's probably better to use `getResources().getInteger(android.R.integer.config_mediumAnimTime)` instead of a hard coded `400`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/searchlist/SearchModule.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.schedjoules.eventdiscovery.framework.searchlist;
+
+/**
+ * A controller-like component that represents a module of a search list. It can react to user search queries by updating it's section in the list.
+ *
+ * @author Gabor Keszthelyi
+ * @see SearchModuleFactory
+ * @see SearchModuleFactory
```
dmfs: That's right, better look twice ;-)

dmfs (changes requested):
Very nice. This much smoother than before. Please don't forget to remove the dummy modules and resubmit for merge.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/LocationSelectionFragment.java
lemonboston: Yes, that's probably better, updated.

lemonboston:
Thanks, I've removed the dummy modules and added a big TODO description for the possible update aggregation. As it seems to work good now, we can also put that aside. Please decide.
Implementing that I think would be 2 to 5 hours work.
Other than that, it is ready for merge from my side.

dmfs (approved):

PR #126, opened 2017-02-22
PR for 95: Current city location picker module

lemonboston:
@dmfs It’s ready for a first review.

TODOs, questions, notes (updated continuously):

- [x] 1. Review and test different Android and targetSdk usage again, also with manifest merge exclusion. (not fully, see item 18.)

- [ ] 2. Update Wiki with this location permission topic. -> update on branch `location-picker/95-current-location` of the wiki repository, needs review and merge

- [x] 3. When the user denies the permission (or had denied the permission earlier) with the "Never ask again" checkbox selected, then we need to hide the Current Location module.

- [x] 4. In order to not fetch the city every time it is shown, i.e. when user clears the search query, I plan to add a simple caching with a few minutes expiration time

- [ ] 5. Translations for new string resources

- [x] 6. In the current way, the last known geo location of the phone is used to fetch corresponding `Address`es with `Geocoder`, and city and country is taken from the first `Address`. The resulting `GeoPlace` will have this geo location as coordinates, so this will not be the same as the one which would have been used from a `AutocompletePrediction`, so when user selects the same city from suggestions.
So the event discovery will have different center of radius.
I've also deliberately added the geo location to the `id()` field of `GeoPlace`, so even two 'current location' city will not be equal if there is any change in last known geo location.
Just thinking now that it may cause problem with 'Recent Location', showing the same city multiple times.
So we probably need to figure this out. Any suggestions?
-> Will be handled with different `Equalable` decorators probably later. Also we may not need to store current location selections in recent locations.

- [x] 7. As I mentioned on chat, I think it would be good to change `GeoPlace` so that it extends `NamedPlace` and `GeoLocation` and not composed of them. What do you think? If you agree, when should we do it? -> good like this

- [x] 8. I've introduced a general onClick action handling for `SmartView`s, please check it out if it looks good to you. In particular, what I don't really like currently that all `ClickeableSmartView`s, in order to implement the `setOnClickAction`, has to store the data, and also that listener setup is repeating over them. See `MessageItemView` for example. -> update as discussed on the call

- [x] 9. A general note, that we are generating many objects in this search list currently. It's partly because of the declarative design around these `change`s and their decorators/adapters, and also because all modules always react to a new query (current location always sends a `Clear` update when query is not empty). There are a couple of classes which we could have a single instance.
A bit unrelated, but the change aggregation will also be still there as an optimization possibility.
So let me know if we should spend any time now to look through the code from this point of view or if you think it's fine. -> not a concern

- [x] 10. Currently after locating the city, it is not selected automatically as in the story, but displayed first, so user has to to select it. I think it’s better this way, it gives a clear feedback that we located him, what city we located and then he can actually confirm by pressing it that it’s right, this is where I am, this is where I want to search. It is also more in line with the general usage of the screen. (We may have nearby places later for example.)
What do you think? -> fine this way

- [x] 11. Review current location permission 'flow' and related user messages. Should we do something differently? Please check starting typing, erasing, and switching permission in phone settings.
Should we maybe add an ‘X’ button to messages displayed after user has denied the permission? I think for the last one that informs them to go to phone settings should definitely have an 'X'.
If we add 'X' to the first denied message, too, the question is whether to show the Current Location module next time.
Should we maybe add it to this error message item as well, to inactivate the module for the time of the current activity?
_“Failed to get the location. Tap to retry.”_

- [x] 12. I’ve added and started to use this `Equalable` interface as mentioned. Your suggestion is a bit different, so we need to get back to this, discuss it, and update.
Separate ad-hoc technical story? -> will be handled it separate story

- [x] 13. Possible generalization of `LocationPermissionProxy`  - I don't know at this point that it's worth it, handling might be special to module, but we should at least do it until 11. is finalized. -> generalized

- [x] 14. TODO at `PlaceSuggestionModule.java:141` for code improvement

- [x] 15. Android `Location` uses `double` for latlng while we use `float` so there is conversion (in both direction). Could it cause any problem when comparing? Considering using `double` on server as well, given that google is used there as well?

- [x] 16. I plan to move `framework.location.model` package to `framework.model.location` and rename `framework.location` to `framework.locationpicker`. Okay to do it now, or postpone it because of possible code conflicts with other story? -> will be listed as new story

- [x] 17. The `FourWayPermissionStatus` now handles the case when the manifest doesn't contain the permission (it was removed) as well. Should we make a `FiveWayPermissionStatus` with an extra `NOT_IN_MANIFEST` status? That case is now mapped to `DENIED_WITH_NEVER_ASK_AGAIN`, which is mostly good, but this 5th one might be useful to know not to ever advise the user to grant the permission from phone settings, because they can't. It is not a big effort to add this. -> updated to `FiveWay` with `NOT_IN_MANIFEST`.

- [x] 18. Testing the app installation permission request on API level < 23 device. I couldn't do it with the emulator, simply can't find the apk file I put on the emulator. It has no file explorer, just 'Downloads' app, that doesn't show it even though I put it in sdcard/dowloads. The file can be seen however from 'Settings/Storage/Misc', but that it can be 'executed'.

- [x] 19. Create story for Equalable stuff -> https://github.com/schedjoules/android-event-discovery-sdk/issues/129

- [x] 20. Create story for UI, UX improvements that are not critical, like animating dots on "Locating..." and animating showing up the found location. (Depends on the review)

- [x] 21. Create story for package refactors, moving classes:
Move `framework.location.model` package to `framework.model.location` and rename `framework.location` to `framework.locationpicker`. Move modules related classes in `framework.locationpicker` to subpackage `modules`. -> created: https://github.com/schedjoules/android-event-discovery-sdk/issues/130

- [x] 22. See if it's worth adding and extra layer above permission broadcast handling. Some of the code in the `BroadcastReceiver` would be repeated in other `Permission`s. See TODO there.
-> I tried a few ways, none of them was totally clear, so I kept it like this for now, if we have a new Permission, we can look at it again

- [x] 23. Possibly create a story to use center of the city as geo location when selecting current location instead of the last known location of the phone. It may be possible to do it relatviely easy with an extra call to `getFromLocationName()` of `GeoCoder`.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/CurrentLocationPermissionProxyFactory.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/CurrentLocationPermissionProxyFactory.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.location;
+
+import android.Manifest;
+
+import com.schedjoules.eventdiscovery.R;
+import com.schedjoules.eventdiscovery.framework.location.model.GeoPlace;
+import com.schedjoules.eventdiscovery.framework.searchlist.SearchModuleFactory;
+import com.schedjoules.eventdiscovery.framework.searchlist.permissionproxy.AbstractPermissionProxyModuleFactory;
+import com.schedjoules.eventdiscovery.framework.searchlist.permissionproxy.PermissionProxy;
+import com.schedjoules.eventdiscovery.framework.searchlist.predicate.EmptyQueryPredicate;
+
+
+/**
+ * {@link SearchModuleFactory} that decorates {@link CurrentLocationModule} with a {@link PermissionProxy} for {@link
+ * Manifest.permission#ACCESS_FINE_LOCATION}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class CurrentLocationPermissionProxyFactory extends AbstractPermissionProxyModuleFactory<GeoPlace>
+{
+
+    public CurrentLocationPermissionProxyFactory(SearchModuleFactory<GeoPlace> delegateModuleFactory)
+    {
+        super(delegateModuleFactory,
+                new EmptyQueryPredicate(),
+                Manifest.permission.ACCESS_FINE_LOCATION,
+                R.string.schedjoules_location_picker_current_location,
+                R.string.schedjoules_location_picker_current_location_permission_rationale,
+                R.string.schedjoules_location_picker_current_location_permission_inform_about_phone_settings);
+    }
+
+}
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/Equalable.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/Equalable.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.model;
+
+/**
+ * Interface for marking types whose implementation has to override equals() and hashCode().
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface Equalable
+{
+    /**
+     * See {@link Object#equals(Object)}
+     */
+    boolean equals(Object other);
+
+    /**
+     * See {@link Object#hashCode()}
+     */
+    int hashCode();
+
+}
--- a/eventdiscovery-sdk/src/main/res/values/dimens.xml
+++ b/eventdiscovery-sdk/src/main/res/values/dimens.xml
@@ -15,5 +15,7 @@
     <dimen name="schedjoules_event_list_progress_bar">36dp</dimen>
     <dimen name="schedjoules_event_detail_item_icon">28dp</dimen>
 
+    <dimen name="schedjoules_location_list_item_height">64dp</dimen>
+
     <dimen name="schedjoules_toolbar_title_font_size">20dp</dimen>
 </resources>
--- a/.idea/dictionaries/shared_for_this_project.xml
+++ b/.idea/dictionaries/shared_for_this_project.xml
@@ -3,6 +3,7 @@
     <words>
       <w>clearable</w>
       <w>eventdiscovery</w>
+      <w>geocoder</w>
       <w>sched</w>
       <w>schedjoules</w>
       <w>sectionable</w>
--- a/eventdiscovery-demo/src/main/AndroidManifest.xml
+++ b/eventdiscovery-demo/src/main/AndroidManifest.xml
@@ -2,6 +2,10 @@
 <manifest package="com.schedjoules.eventdiscovery.demo"
           xmlns:android="http://schemas.android.com/apk/res/android">
 
+    <!-- Add this to your manifest if your targetSdkVersion < 23: -->
+    <!--suppress AndroidDomInspection -->
+    <!--<uses-permission-sdk-23 tools:node="removeAll"/>-->
+
     <application
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
--- a/eventdiscovery-sdk/build.gradle
+++ b/eventdiscovery-sdk/build.gradle
@@ -66,6 +66,7 @@ dependencies {
     compile 'com.github.bumptech.glide:glide:3.7.0'
     compile 'net.opacapp:multiline-collapsingtoolbar:1.3.0'
     compile 'com.google.android.gms:play-services-places:9.8.0'
+    compile 'com.google.android.gms:play-services-location:9.8.0'
     compile('eu.davidea:flexible-adapter:5.0.0-rc1') {
         exclude group: 'com.android.support'
     }
--- a/eventdiscovery-sdk/src/main/AndroidManifest.xml
+++ b/eventdiscovery-sdk/src/main/AndroidManifest.xml
@@ -2,6 +2,7 @@
           package="com.schedjoules.eventdiscovery">
 
     <uses-permission android:name="android.permission.INTERNET"/>
+    <uses-permission-sdk-23 android:name="android.permission.ACCESS_FINE_LOCATION"/>
 
     <application>
 
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/async/CachingSafeAsyncTask.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/async/CachingSafeAsyncTask.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.async;
+
+import android.support.annotation.WorkerThread;
+
+import com.schedjoules.eventdiscovery.framework.model.Equalable;
+import com.schedjoules.eventdiscovery.framework.utils.cache.Cache;
+
+
+/**
+ * {@link SafeAsyncTask} that uses a {@link Cache} to get results from if they are cached, otherwise executes that task and saves the result in the cache.
+ *
+ * @author Gabor Keszthelyi
+ */
+public abstract class CachingSafeAsyncTask<TASK_PARAM extends Equalable, EXECUTE_PARAM, PROGRESS, TASK_RESULT> extends SafeAsyncTask<TASK_PARAM, EXECUTE_PARAM, PROGRESS, TASK_RESULT>
+{
+    private final Cache<TASK_PARAM, TASK_RESULT> mCache;
+
+
+    public CachingSafeAsyncTask(TASK_PARAM taskParam, SafeAsyncTaskCallback<TASK_PARAM, TASK_RESULT> callback, Cache<TASK_PARAM, TASK_RESULT> cache)
+    {
+        super(taskParam, callback);
+        mCache = cache;
+    }
+
+
+    @Override
+    protected void onPreExecuteWithParam(TASK_PARAM taskParam)
+    {
+        TASK_RESULT cachedValue = mCache.get(taskParam);
+        if (cachedValue != null)
+        {
+            cancel(false);
+            onPostExecute(new SuccessTaskResult<>(cachedValue));
+        }
+    }
+
+
+    @Override
+    protected final TASK_RESULT doInBackgroundWithException(TASK_PARAM taskParam, EXECUTE_PARAM... execute_params) throws Exception
+    {
+        TASK_RESULT cachedValue = mCache.get(taskParam);
+        if (cachedValue != null)
+        {
+            return cachedValue;
+        }
+        TASK_RESULT taskResult = doInBackgroundWithExceptionForNonCached(taskParam, execute_params[0]);
+        mCache.put(taskParam, taskResult);
+        return taskResult;
+    }
+
+
+    @WorkerThread
+    protected abstract TASK_RESULT doInBackgroundWithExceptionForNonCached(TASK_PARAM taskParam, EXECUTE_PARAM execute_param) throws Exception;
+}
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/async/SafeAsyncTask.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/async/SafeAsyncTask.java
@@ -38,7 +38,20 @@ public abstract class SafeAsyncTask<TASK_PARAM, EXECUTE_PARAM, PROGRESS, TASK_RE
     public SafeAsyncTask(TASK_PARAM taskParam, SafeAsyncTaskCallback<TASK_PARAM, TASK_RESULT> callback)
     {
         mTaskParam = taskParam;
-        mCallback = new WeakReference<SafeAsyncTaskCallback<TASK_PARAM, TASK_RESULT>>(callback);
+        mCallback = new WeakReference<>(callback);
+    }
+
+
+    @Override
+    protected final void onPreExecute()
+    {
+        onPreExecuteWithParam(mTaskParam);
+    }
+
+
+    protected void onPreExecuteWithParam(TASK_PARAM taskParam)
+    {
+
     }
 
 
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/http/RequestUriLogging.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/http/RequestUriLogging.java
@@ -35,7 +35,7 @@ import java.net.URI;
  *
  * @author Gabor Keszthelyi
  */
-public class RequestUriLogging implements HttpRequestExecutor
+public final class RequestUriLogging implements HttpRequestExecutor
 {
 
     private final HttpRequestExecutor mDelegate;
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/ItemChosenAction.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/ItemChosenAction.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.list;
+
+/**
+ * Action that can be executed when a list item was chosen by the user.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface ItemChosenAction<T>
+{
+
+    /**
+     * Called when the item was chosen by the user.
+     *
+     * @param itemData
+     *         the data corresponding to the list item
+     */
+    void onItemChosen(T itemData);
+}
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/changes/nonnotifying/ClearAll.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/changes/nonnotifying/ClearAll.java
@@ -27,6 +27,7 @@ import java.util.List;
  */
 public final class ClearAll<T> implements NonNotifyingListChange<T>
 {
+
     @Override
     public void apply(List<T> currentItems)
     {
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/changes/nonnotifying/ReplaceAll.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/changes/nonnotifying/ReplaceAll.java
@@ -17,6 +17,7 @@
 
 package com.schedjoules.eventdiscovery.framework.list.changes.nonnotifying;
 
+import java.util.Collections;
 import java.util.List;
 
 
@@ -33,6 +34,12 @@ public final class ReplaceAll<T> implements NonNotifyingListChange<T>
     public ReplaceAll(List<T> newItems)
     {
         mNewItems = newItems;
+    }
+
+
+    public ReplaceAll(T item)
+    {
+        this(Collections.<T>singletonList(item));
     }
 
 
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/smart/AbstractSmartListItem.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/smart/AbstractSmartListItem.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.list.smart;
+
+import android.support.annotation.LayoutRes;
+import android.view.View;
+
+import com.schedjoules.eventdiscovery.framework.list.ListItem;
+import com.schedjoules.eventdiscovery.framework.utils.smartview.SmartView;
+
+
+/**
+ * Abstract class for {@link ListItem}s that work with {@link SmartView}s.
+ * <p>
+ * Note: Usually list items' 'equality' should be based on the data/model objects (type D) equality, so make sure that the data type equality is defined.
+ *
+ * @author Gabor Keszthelyi
+ */
+public abstract class AbstractSmartListItem<D, V extends View & SmartView<D>> implements ListItem<V>
+{
+    private final D mItemData;
+    private final int mLayout;
+
+
+    public AbstractSmartListItem(D itemData, @LayoutRes int layout)
+    {
+        mItemData = itemData;
+        mLayout = layout;
+    }
+
+
+    @Override
+    public final int layoutResId()
+    {
+        return mLayout;
+    }
+
+
+    @Override
+    public final void bindDataTo(V view)
+    {
+        view.update(mItemData);
+    }
+
+
+    @Override
+    public final boolean equals(Object o)
+    {
+        if (this == o)
+        {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass())
+        {
+            return false;
+        }
+
+        AbstractSmartListItem<?, ?> that = (AbstractSmartListItem<?, ?>) o;
+
+        return mItemData.equals(that.mItemData);
+
+    }
+
+
+    @Override
+    public final int hashCode()
+    {
+        return mItemData.hashCode();
+    }
+
+
+    @Override
+    public final String toString()
+    {
+        return toStringLabel() + "{" +
+                "itemData=" + mItemData +
+                '}';
+    }
+
+
+    /**
+     * Label that is added to the toString() value of the object, practically use the class name in string literal, like "MyListItem".
+     */
+    protected abstract String toStringLabel();
+}
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/smart/Clickable.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/smart/Clickable.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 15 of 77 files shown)


lemonboston:
I faced an issue with the permission remove for the manifest merge: it didn't work for our `uses-permission-sdk-23`. After about an hour and a half troubleshooting, I found why:
https://developer.android.com/studio/build/manifest-merge.html#appendix_merge_policies
Table 3 lists the supported elements and `uses-permission-sdk-23` is not among them.
And for custom elements:
_No matching; these are unknown to the merger tool so they are always included in the merged manifest_

I've pushed it to branch `location-picker/95-current-location-permission-issue`, so you can check it out as well.

So without a better solution, I propose now that we don't add anything to the manifest in the SDK, but instruct partners to add it on their side with `uses-permission-sdk-23` if they have `targetSdk` >= 23. 
What do you say?

lemonboston:
Caching the current city added with 3 minutes expiry time.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/smart/AbstractClickableSmartListItem.java
```diff
@@ -0,0 +1,123 @@
...
+
+
+/**
+ * Abstract class for {@link ListItem}s that work with {@link ClickableSmartView}s.
+ * <p>
+ * Note: Usually list items' 'equality' should be based on the data/model objects (type D) equality, so make sure that the data type equality is defined.
+ *
+ * @author Gabor Keszthelyi
+ */
+public abstract class AbstractClickableSmartListItem<D, V extends View & ClickableSmartView<D>> implements ListItem<V>, SmartClickable<D>
```
dmfs: Do we really need this class? Can't we just have a `Clickable` decorator? All it would have to do is set its own click handler in `bindDataTo()` or am I missing something?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/smart/AbstractClickableSmartListItem.java
```diff
@@ -0,0 +1,123 @@
...
+{
+    private final D mItemData;
+    private final int mLayout;
+
+    // Note: In some circumstances it is useful if the action can be set after the item has been created, that's why it's not final.
```
dmfs: Under which circumstances is that the case?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/LocationPermissionProxy.java
```diff
@@ -0,0 +1,151 @@
...
+import com.schedjoules.eventdiscovery.framework.utils.strings.Strings;
+
+
+/**
+ * A {@link SearchModule} control proxy to use before {@link CurrentLocationModule} that handles the Location permission, i.e. doesn't let queries through until
+ * the permission is granted.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class LocationPermissionProxy implements SearchModule, SinglePermission.PermissionGrantListener
```
dmfs: I like the idea to handle this in a decorator. Though, can you make this more generic, i.e. pass the required permission to the constructor? Other modules might need other permissions (e.g. read calendar).

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/model/EqualableNamedPlace.java
```diff
@@ -0,0 +1,84 @@
...
+
+import com.schedjoules.eventdiscovery.framework.model.Equalable;
+
+
+/**
+ * {@link Equalable} decorator for {@link NamedPlace}, value is based on id().
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class EqualableNamedPlace implements NamedPlace, com.schedjoules.eventdiscovery.framework.model.Equalable
```
dmfs: Please stick to our naming scheme for decorators. This should be just `Equalable`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/Equalable.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.schedjoules.eventdiscovery.framework.model;
+
+/**
+ * Interface for marking types whose implementation has to override equals() and hashCode().
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface Equalable
```
dmfs, line 25: I think we should distinguish two cases:
* calling equal to explicitly compare it to another object of the same type
* making it hashable (having `equals` called by a `Set` or `Map`).

In the first case the `Equalable` interface should be generic to specify the type you can compare to (just like the `Comparable` interface does it). To avoid conflicts with `equals` the method should have a different signature in this case.

eventdiscovery-sdk/src/main/res/values/dimens.xml
```diff
@@ -14,5 +14,7 @@
     <dimen name="schedjoules_event_list_progress_bar">36dp</dimen>
     <dimen name="schedjoules_event_detail_item_icon">28dp</dimen>
 
+    <dimen name="schedjoules_location_list_item_height">62dp</dimen>
```
dmfs: Why 62 and not standard 64?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/SinglePermission.java
```diff
@@ -0,0 +1,55 @@
...
+
+import android.Manifest;
+
+
+/**
+ * Represents a single Android permission. Like {@link Manifest.permission#ACCESS_FINE_LOCATION}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface SinglePermission
```
dmfs: Why "Single..."? Wouldn't it be sufficient to call it just `Permission` or maybe `PermissionGrant`?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/SinglePermission.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    boolean granted();
+
+    /**
+     * Requests the permission from the user.
+     *
+     * @param listener
+     *         listener for notifying about the user's decision
+     */
+    void requestGrant(PermissionGrantListener listener);
```
dmfs: The Javadoc says it requests a "Permission", the name says it requests a "Grant"? I think it's sane to call this method just `request`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/SinglePermission.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface SinglePermission
+{
+
+    /**
+     * Tells whether the permission is granted currently.
+     */
+    boolean granted();
```
dmfs: for consistency we should call methods like this "is...", i.e. `isGranted()`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/SinglePermissionGrantResult.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package com.schedjoules.eventdiscovery.framework.permission;
+
+/**
+ * Represents the result of the user's decision on a single permission request.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface SinglePermissionGrantResult
```
dmfs: How is that different from `SinglePermission`? I guess `SinglePermission.granted()` always returns exactly the same result, doesn't it?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/PermissionGrantResult.java
```diff
@@ -0,0 +1,39 @@
...
+ * Represents the result of the user's decision on a permission request.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface PermissionGrantResult
+{
+    /**
+     * The list of permissions granted by the user.
+     */
+    List<String> granted();
```
dmfs: Does it really make sense to return a list here? Wouldn't it be sufficient to have a single method like `boolean isGranted(String)`?

dmfs (changes requested):
Done with a first round. There is probably more to come. I'll try to test the permission issue next week by uploading a new app to Google Play (unfortunately they don't allow to downgrade the targetSdk once an app has been published with 23+).

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/LocationPermissionProxy.java
```diff
@@ -0,0 +1,151 @@
...
+     * TODO If user denies or had denied at any time the permission with "Never ask again" checked,
+     * we should not show any item (the module basically).
+     */
+
+
+    @Override
+    public void onPermissionGrantResult(boolean granted)
+    {
+        mGranted = granted;
+        if (mGranted)
```
dmfs: This should not depend on the granted state. Ideally this is determined by [shouldShowRequestPermissionRationale](https://developer.android.com/reference/android/support/v4/app/ActivityCompat.html#shouldShowRequestPermissionRationale(android.app.Activity,%20java.lang.String)), also see this comment: http://stackoverflow.com/questions/30719047/android-m-check-runtime-permission-how-to-determine-if-the-user-checked-nev#comment53800740_31925748

lemonboston:
Thanks for the first review. Just some quick notes now, I will get back to all of them above later.

Why do you need Play Store to test the permission issue? I simply used the demo app for that, changed it's `targetSdk`, copy its apk to the phone and install, it will ask for the permission that way (not sure if it can be done from Android Studio).

Yesterday, I ran into the problem of how to determine whether user checked 'Never ask again', and as you've also seen probably, Google missed to make it easy unfortunately. The only somewhat reliable way looks like tracking `onRequestPermissionsResult`, checking `shouldShowRationale` there (http://stackoverflow.com/a/35495893/4247460). So I plan to do something like this:
http://stackoverflow.com/a/35013138/4247460 with using SharedPreferences as well to separate the BLOCKED and NEVER_ASKED state (and maybe 1 extra state, I don't remember now), calling into the 'util' from `onRequestPermissionsResult` to store what's needed. 
I've briefly looked into some libs I found here:
https://android-arsenal.com/tag/235
But none of them seemed like to be what I was looking for, namely to track the detailed status of a permission, so the plan is to implement it on our own. It may be better also not to depend on something we don't know the internals of.
I plan to handle (reasonably if perfectly cannot) and test these cases as well, besides the basic ones:
1. user granted/denied permission in the app, but not on our SDK screen, so no tracking happened in `onRequestPermissionResult`
2. user granted/denied permission from phone settings
3. user deleted app data, so we lost the SharedPref entry

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/smart/AbstractClickableSmartListItem.java
lemonboston: No, we don't! :) Thanks, I've updated it, much nicer now as a decorator.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/list/smart/AbstractClickableSmartListItem.java
lemonboston: This has been removed, so no such usage currently, but if the `ListItem`s are prepared on background thread, then we may need to set the listeners after they have been created. That is the case in the Event list currently, but those are not smart yet.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/LocationPermissionProxy.java
lemonboston: I think we should get back to this at the end, when we finalized this UX flow, to see if it's worth generalizing, passing in all the messages. (Checklist item 13.)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/model/EqualableNamedPlace.java
lemonboston: Updated.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/Equalable.java
lemonboston: Continue this on Slack..

eventdiscovery-sdk/src/main/res/values/dimens.xml
lemonboston: Updated to 64dp, also corrected the layout to align the divider correctly to the bottom.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/SinglePermission.java
lemonboston: Yes, it's probably better, so now there is only `Permission`, and multi permission requests are not supported. Those could have a different interface, but it's not needed at the moment.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/SinglePermission.java
lemonboston: Since this has moved from `Permission` to new `PermissionStatus`, it felt better to keep it as `granted()`, let me know if I should change it nevertheless.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/SinglePermission.java
lemonboston: Right, I agree, I used just `request()` for the updated interface.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/SinglePermissionGrantResult.java
lemonboston: This has been removed.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/PermissionGrantResult.java
lemonboston: This has been removed.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/LocationPermissionProxy.java
lemonboston: See more about this in the main thread..

lemonboston:
@dmfs This is ready for a 2nd round of review now.

There are several open questions left, I've updated the checklist at the first comment above, please check them as well.

About the permission design:
So there was the need to distinguish between _never asked_ and _denied with checkbox_, and to the best of my knowledge now, it can only be tracked by checking the `shouldShowRationale` boolean upon `onRequestPermissionResult()` and store it.
But if we start to store something in `SharePref`, I thought it's more straightforward if the current status is stored regardless of which one is that.
But this is just one way to represent permission statuses, so I created the interface for `PermissionStatus` and used an enum for this `FourWay` one. This was a nice revelation for me that enums can implement an interface, by the way :), and that I can generlize it this way.
For the universal broadcast sent from `onRequestPermissionResult()`, I didn't want it to be interpreted, or abstracted, so that's why the `PermissionRequestResult` have those `public final` fields.
But I've seen since, that it may be worth putting a layer/abstraction above these permission broadcast, because some of the code in the receiver would be repeated (added a TODO there). So this could be possibly improved further a bit, although doesn't seem to be important.

lemonboston:
Since the Wiki is not under version control, I put here a proposal for change when this pull request is merged:

Under the 'Add internet permission' section:
https://github.com/schedjoules/android-event-discovery-sdk/wiki/5-minutes-install#add-internet-permission

## Only if your `targetSdkVersion` is lower than 23, add this:

```
    <!--suppress AndroidDomInspection -->
    <uses-permission-sdk-23 tools:node="removeAll"/>
```
This removes the `ACCESS_FINE_LOCATION` permission that the SDK normally requests runtime at the place of usage on the location picker screen, but with `targetSdkVersion` < 23, the permission would be requested upon installation and to avoid that you can remove the permission definition this way. (This of course also means the current location detection will not be available in the SDK)

lemonboston:
@dmfs It's ready for review now, I don't have any more planned changes or open questions.
Please also check the not completed items on the checklist above.

lemonboston:
Removed `Activity` parameter from `OnClickAction.onClick()` and squash-rebased to master.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/Permission.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.schedjoules.eventdiscovery.framework.permission;
+
+/**
+ * Represents a single Android permission.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface Permission<S extends PermissionStatus>
```
dmfs: Why is `Permission` generic? Under what condition would I want a `Permission` of a different `PermissionStatus`?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/PermissionRequestCallback.java
```diff
@@ -0,0 +1,40 @@
...
+public interface PermissionRequestCallback<S extends PermissionStatus>
+{
+
+    /**
+     * Called when the user answered the permission dialog.
+     *
+     * @param newStatus
+     *         the new status of {@link Permission}
+     */
+    void onUserAnswered(S newStatus);
```
dmfs: I think these method names are a bit too verbose. `onResult` and `onInterrupt` should be sufficient.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/broadcast/PermissionRequestResult.java
```diff
@@ -0,0 +1,81 @@
...
+
+
+/**
+ * The data for the {@link Activity#onRequestPermissionsResult(int, String[], int[])} callback.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class PermissionRequestResult implements Parcelable
+{
+    public final int requestCode;
```
dmfs: public fields? Sorry, that's a no-go. This class doesn't seem to have a real purpose except for one-way-parcelling. Since it doesn't implement any other interface it actually doesn't give access to the parcelled data.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/fiveway/FiveWayPermission.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private static final int REQUEST_CODE = 45417;
+
+    private final Activity mActivity;
+    private final String mPermissionName;
+
+
+    public FiveWayPermission(Activity activity, String permissionName)
+    {
+        mActivity = activity;
```
dmfs: Have you considered to make `Activity` or `Context` a parameter to `status()` and `request()` instead?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/fiveway/FiveWayPermissionTracker.java
```diff
@@ -0,0 +1,246 @@
...
+    public FiveWayPermissionStatus checkStatus(Activity activity, String permissionName)
+    {
+        return checkStatus(activity, permissionName, true, LOG_ENABLED);
+    }
+
+
+    @Override
+    public TrackablePermissionRequestResult trackableResult(final PermissionRequestResult result)
+    {
+        return new TrackablePermissionRequestResult()
```
dmfs: What's the purpose of this `TrackablePermissionRequestResult`? Why this indirection?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/tracker/PermissionTracker.java
```diff
@@ -0,0 +1,51 @@
...
+ * Because the state check supported by Android is limited, in order to have more detailed status, implementation might need to store values, and register
+ * {@link Activity#onRequestPermissionsResult(int, String[], int[])} results with the should show rationale flag at that moment. See <a
+ * href="http://stackoverflow.com/q/30719047/4247460">Related StackOverflow discussion</a>
+ * <p>
+ * The aim of this interface is to have those coupled logic - the checking of the status and the registering of the permission results - together. But it should
+ * be used under a {@link Permission}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface PermissionTracker<S extends PermissionStatus>
```
dmfs: Same here, why would I want a `PermissionTracker` of a different type?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/CurrentLocationPermissionProxyFactory.java
```diff
@@ -0,0 +1,59 @@
...
+        super(delegateModuleFactory,
+                new EmptyQueryPredicate(),
+                Manifest.permission.ACCESS_FINE_LOCATION,
+                R.string.schedjoules_location_picker_current_location,
+                R.string.schedjoules_location_picker_current_location_permission_rationale,
+                R.string.schedjoules_location_picker_current_location_permission_inform_about_phone_settings);
+    }
+
+
+    private static class EmptyQueryPredicate implements QueryPredicate
```
dmfs: This class doesn't need to be private.

dmfs (changes requested):

dmfs:
Forgot to mention the activity leakage in the review ...

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/Permission.java
lemonboston: `PermissionStatus` is generalized to allow multiple ways to define statuses. It's probably true that within one app we won't use different interpretations, but this 5-way version is still an opinionated version (the fact that it was 4-way originally shows this as well), so I wanted to keep the option to use other versions. The trivial would be the 2-way version which only checks for current `checkSelfPermission()` result, so basically what Android provides. And in some circumstance that might be enough, so no need to handle 5 cases of enum values. In that case, no broadcasting is needed in the app.
(The 5-way version to function best assumes of course that it is used always for a certain permission, to be able to track its status properly. Still that wouldn't mean I think that we should restrict other status usage if it fits better somewhere)

And of course `Permission` is generic because of `PermissionStatus`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/PermissionRequestCallback.java
lemonboston: ok, changed it

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/broadcast/PermissionRequestResult.java
lemonboston: This was deliberately intended as what you call a data bag. The data that was delivered in `Activity.onPermissionRequestResult(int, String[], int[])`, to be able to send them in a broadcast without any interpretation. I didn't want to add any abstraction to this part. I wanted it to be independent from `Permission` and `PermissionStatus`es, and allow any interested party to listen in on this broadcast, like the partner app itself, if it's something useful for them.
So this is intended to be raw data, sort of. Not opinionated, not generalized.
And if it's just raw data, I am not sure that there will be any use for the interface, maybe there will be, I just don't see it.

So if we want to change this, I see 3 options now:

1. Create interface with the same methods for the raw data:
```
interface PermissionResult
{
     int requestCode();
     String[] permissions();
     int[] grantResults();
}
```

2. Add them separately to the broadcast Intent with 3 keys

3. Create some interpreted interface to represent the result, not sure how best to do it, 2 examples:

a.
```
interface PermissionResult
{
     boolean wasInterrupted();
     boolean granted(String permissionName) throws IllegalArgumentExcepttion, IllegalStateException;
}
```
b.
```
interface PermissionResult
{
     boolean wasInterrupted();
    List<String> granted();
    List<String> denied();
}
```

But as I mentioned I would avoid adding another abstraction 'layer' here, so I would go with 1., or maybe 2.
What do you say?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/fiveway/FiveWayPermission.java
lemonboston: I've looked into it now and one benefit of moving them to parameters could be that `status()` only needs `Context` normally, so it could be used from a `Service` as well, but unfortunately `FiveWay` implementation uses `Activity` for `status()` too, to check for the `shouldShowRequestPermissionRationale` state. So what I can do is this:

```
interface Permission<S extends PermissionStatus, C extends Context>
{
    S status(C context);
    void request(Activity activity, PermissionRequestCallback<S> callback);
}

class FiveWayPermission implements Permission<FiveWayPermissionStatus, Activity>
```
So `FiveWayPermission` could only be used from `Activity` context.
This change would also mean we need to save `Activity/Context` references in components where we want to use `Permission`, like in current `PermissionProxy`.
But the possibility to use a `Permission` from non-Activity context, I think justifies this change, so I say we should change it as above. What do you say?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/broadcast/PermissionRequestResult.java
dmfs: I wouldn't call it abstraction layer. Call it proper design. A `PermissionResult` sounds like a well defined entity. From how you use it I'd say that something like 
```
interface PermissionResult
{
    Iterable<String> granted();
    Iterable<String> denied();
}
```

Should be almost sufficient. Don't return a list unless you need to access elements by index (i.e. you're actually operating on a list). Returning a list makes no sense here. A set would be more appropriate, but even that doesn't really make sense, because you'd barely check for a specific permission to be in here.

I understand that the interruption of such a request is a rare (error) condition, hence it might not make sense at all to have a separate method to check it. Just throw an appropriate Exception in `granted` and `denied` when called.
Or is there any valid case in which the request can be interrupted (anything but an error)?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/fiveway/FiveWayPermissionTracker.java
lemonboston: I wanted to avoid procedural here, not sure if it makes sense. Similar to that `RecentLocation.remember(), forget()` thing.
If you don't see any benefit of that here, I remove it and change it to `void trackResult(activity, result)`. Please let me know.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/tracker/PermissionTracker.java
lemonboston: This is still about `PermissionStatus` being general, so I think https://github.com/schedjoules/android-event-discovery-sdk/pull/126#discussion_r104903698 answers that.
Although this whole `PermissionTracker` interface may not be needed, but then the `FiveWayPermission` implementation becomes too big. So it basically serves that purpose to extract the 'tracking' logic/interpretation corresponding to a `PermissionStatus` from a `Permission`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/CurrentLocationPermissionProxyFactory.java
lemonboston: Made it a top level class.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/permission/fiveway/FiveWayPermissionTracker.java
dmfs: In `Recent<T> Recents.recent(T value)` I'm using this pattern because the actual `Recent` implementation depends on `Recents` and only `Recents` can create the appropriate class.

I can't see any reason for the same pattern in FiveWayPermissionTracker.

lemonboston:
Squashed and rebased to master after the new Permission package had been merged.

lemonboston:
I've just realized that the used `GeoCoder` can geo code in the other way as well, so we may be able to find the geo for the center of the city with an extra call to `getFromLocationName()`.
https://developer.android.com/reference/android/location/Geocoder.html#getFromLocationName(java.lang.String,%20int,%20double,%20double,%20double,%20double)
I add a checklist item on the top to possibly create an improvement story for it.

lemonboston:
I've just checked the update to `Permission`s with the 2 methods now `isRequestable()` and `isGrantable()`. It's still hard for me to get a quick and easy understanding in my mind about what do they return in which cases. So for example I don't understand why `isRequestable()` can return `true` for <23 levels, while we cannot actually request anything at runtime there.
The two methods starts with the same description:
"Returns whether this permission can be requested at runtime"
While I understand the point of being able to handle permissions uniformally before and after level 23, I still have a feeling that this ambiguities of meanings make conceptually harder to work with them, and that unnecessasry usage of `Permissioin` below level 23 with decorators can also add further understandability issues and possible bugs as well.
I think the fact that some of these methods will always return true or false or a value of another method, or a fake object, under level 23 may indicate that those should not be part of the model on those levels.
But I don't object going on with the current state.

I don't have anything pending on my side for this story, @dmfs please feel free to act on the remaining checklist items, create separate stories and merge as you see fit.

dmfs:
Think of the permissions package as an emulator of Android 6 permissions on older platforms (it just never actually prompts the user). The main advantage is that you never ever have to perform a check like `Build.VERSION.SDK_INT >= 23` with this library when it comes to permissions. Handling permissions on all platform versions is important for a library like ours, because the actual permissions in the manifest are not under our control. Which means, even on Android 5 and below, we may have check for permissions that we added to the manifest of our library, because partners can remove or not add them.
I think that being able to write platform independent code is a real treat. Ideally our code would never ever have to check on which platform version it's running. That should be handled by abstraction layers like this one.
Regarding the methods that always return `true` or `false`: that's definitely not an indicator to remove them. That's just part of the abstraction/emulation layer. It's a platform specific implementation of the model. On platform versions >=23 these methods do some actual work. Having these methods allows you to remove a number of `Build.VERSION.SDK_INT >= 23` checks before you work with permissions.

Regarding `isRequestable` returning `true` on older platforms. It's true that (using Android methods) you can not request runtime permissions on older platforms. However it only returns `true` for permissions that you already have. Which means you most likely won't even call it in that case. And if you call it and fire a permission request the "emulation layer" just fakes a granted permission. The only flaw in that is that `onRequestPermissionsResult` is never called. But IMO you can and *should* put all handling of permissions in `onPause` anyway, because permissions can be changed without `onRequestPermissionsResult` being called (e.g. the user presses home, goes to the settings and changes your app permissions and returns via the "recent apps"). If you only act on `onRequestPermissionsResult`  you will miss these changes.

That being said, I certainly need to update the javadoc. Apparently I missed that one.

lemonboston:
Ok, about the case of changing permission in phone settings, I think the system may reload the Activity in that case, at least that how it looked like to me. I will double-check this today.

lemonboston:
I just checked what happens when changing permission from phone settings, it's different between denying and granting:
- when change to denied, the system kills the process of the app and restores the Activity when going back to recent apps
- when changed to granted, nothing like that happens, normal resuming
So I suppose refreshing in `onResume()` is still useful because of this second case.

dmfs (approved):

dmfs:
@lemonboston I've made a few more changes. After adding translations, please squash and resubmit. Everything else well be subject to other stories or bug reports.

dmfs:
@lemonboston please add translations in #150 and squash

lemonboston:
Translations added and squashed.

Issue #95, opened 2017-01-20
In order to search events at my current location, as a user I want an option to "pick" my "current location" in the LoPi.


dmfs:
The behavior is expected to be as follows:
* if visible, the "current location" module is always the topmost item
* the "current location" item has no section header
* on Android 5.x and lower the "current location" item is only visible if the app has location permission (either coarse or GPS)
* on Android 6 the item is always visible
* if the permission has already been granted the item contains the current location (The name of the city and country should be sufficient) and clicking the item will accept it as the result
* if the permission has not been granted, the title is "Current location" and clicking it will open the permission consent screen. If the permission has been granted the location picker will close and use the current location without any addition click. Otherwise it will show a message like "To use your current location, please grant permission to access your location." and return to the picker.
* If the permission has been granted but the location is not known yet (i.e. not GPS fix yet), we show a subtle indeterminate progress animation ("locating you ...")

lemonboston:
ETA for this story was end of today (Wednesday) but it will be delayed. It took longer to finish the changes and prepare the code for the first pull request #126 . And there are still things to implement and discuss (see in PR).
New ETA depends on some of those questions, so a new, conservative estimation is end of Friday.

PR #142, opened 2017-03-07
Stories/100 recent locations

dmfs:
Ready for a first review.
Known issues: 
* recents don't get purged automatically yet
* sections titles are missing


lemonboston:
It's great to see the recent selections showing up in the picker :)

A few notes, questions so far (I plan to get back to it tomorrow and continue):

`LocationSelectionFragment`
I suppose saving the recent in `ListItemSelectionAction` could be a decorator.

`Recent`
I was wondering whether weight should be here or be an external thing and only a timestamp provided here. I guess weight will mostly be based on timestamp and `T`. If we want to define or modify with decorator in any other way, not sure how could we.. 
But in general, assigning weight to a `Recent` could be an independent thing.
Maybe we could have a descriptor attached next to the value in a `Recent`, something that holds anything relevant about the context when the `Recent` was remembered. So another generic parameter.
In our case it's simply a timestamp, so a `Long` or `DateTime` would do it.
But in other cases it might be something more custom, like if we wanted to show recently looked at Events. Then we may want to order them based on where they were opened from, that could be part of the context, or may want to have different displaying orders used for whatever reason.
So we would have the `T value()` and `C context()` for a `Recent`, and we encapsulate everything that might be relevant later in `C`.
And then ordering becomes an external thing, similar to the `Comparable` in `Prioritized` now, but with `C` context. We could order the same history in different ways.

Currently weight 'calculation' is I think entangled with serialization details in `SharedPreferencesRecents`. So its like `TimestampWeightedSharedPreferencesRecents`.
As I see it what would be needed inside a general `SharePreferencesRecents` are two things:
`T <-> SharePref`
`T, timestamp -> weight`
So these could be its dependencies.
Using `CharSequenceFactory` is one way to implement T <-> SharePref.

As I see in the current state `Recents` are not queried and showed when we type. I think it would be great to have that. So we can show the last 2 recents, let's say, on screen load below current location, but then when user starts to type, we show 1 or 2 from recents that matches, on top. Nice way to find something again with 1 or 2 letters entered.
In this case though, it might be more needed to have that 'X' button as well to remove from history, so the same items always showing up don't annoy the user forever. Do you plan to add that 'X' button?
I guess it could be a separate story as well, together with this querying.

There are several classes that could be used from the current location branch, are you planning to wait for merging that and rebasing, or merge this if finished first?

For the title item, there is `TextHeaderItem` there that I used previously, simply adding it to the `List<ListItem>` on each update. I guess that could be a decoration as well for some of that related interface (`ListChange`, `ResultUpdate`, `NonNotifyingChangeAbleList`).

lemonboston:
I forgot 2 notes:

I was wondering whether it would be good to persist Recents in a way that can be retrieved in order right away, and it could be if all is put into one String entry. Probably not important.

Related to what I mentioned that `CharSequenceConverter` is one way to implement putting and extracting `T` to/from `SharedPrefs`, there is already another way in the code base in `SharedPrefLastSelectedPlace`. 
I am still working on the `Serialzation<V, Container>` concept branch to figure out some details, I think that would provide the exact abstraction that would fit in here (with a `Serialization<V,SharedPreferences>`).

dmfs:
Good points. I've incorporated some most of them.

With the current design it's probably difficult to create an `Entitled` decorator that adds a section title. Mostly because the modules send change sets. Nevertheless I like the idea.

I also changed recent to return a timestamp rather than a weight.

The current implementation still has a few flaws, but I think we don't have to address them here. Instead we can improve them later on when we have a clearer idea of what should go how to the recent locations.
Some of the known issues are:
* changing the language will add most places a second time to the recents (in the new language)
* "current location" will go to the recents as well (even if the same city is already in there).

Ideally we use some sort of unique key to identify duplicates (i.e. a hash) but that will also require changes in how the recent places are stored, because not all properties of a `GeoPlace` should source that key. Maybe we use only the geo location and round it to 100 meters or so.

dmfs:
@lemonboston please review

lemonboston (commented):
Just a note that if you notice, too, that the animations after clearing the query are sometimes done in two steps, it is because `PlaceSuggestionsModule` doesn’t clear on empty string right away, but fires the request, so takes more than 10 ms to react, so it is delayed then for 400 ms. It is already fixed on the error handling branch.

`RecentLocationsModule`
- javadoc stayed copy-pasted
- The number of recents as a constructor param?
- `ResultUpdate` classes and `Clickable` could be used from current location branch later
 
Why did you create an `Uri` from `GeoPlace` in `GeoPlaceCharSequenceConverter`? Do you see any potential usage of that or just used it as a util kind of thing that can help serialize to `CharSequence` and parse back?
Could we move the param constants to a public place since they are used elsewhere as well? It’s not clear who should own this I think, so in that case I think it can be better to move to a let’s say `Params` class in the same package.

`new SharedPreferencesRecents<>(activity, "places", new GeoPlaceCharSequenceConverter())`
This `”places”` is hardcoded at 2 places, somehow I don’t really like this, same with nested bundle keys. They can be spread across the code base and they need to be kept in sync. Can we set a constant somewhere?
Using the same literals at multiple places to avoid depending on something external and act like it’s internal seems a little bit like a cheat to me. What do you think?

`SerializableCharSequenceConverter` is not used currently besides the unit test. What is it for, do you have a plan for it?

Regarding saving current location selection - if we don’t want that we can move the `Remembering` decorator application from the `Fragment` to the individual `SearchModuleFactory`s, leaving out the `CurrentLocationModuleFactory`. I suppose currently that would be better.

About the equality of GeoPlaces, I think it would be good to aim for defining it as a `Equalable` decorator. If you don’t mention it I wouldn’t have even thought about how does it work currently  because it’s hidden in `GeoPlaceCharSequenceConverter` and in `SharedPreferencesRecents` internals, so not clearly defined at one place.
Regarding the language duplicates, the `id()` should be the same for them if they are selected from suggestions, that’s the Google id of the place. So if equality is based on the id, it can work. More recent one could override the earlier.
If we want to compare current location selection as well, I think we should probably simply check for name() and extraContext() equality. We probably don’t want to show the same item twice even if they have different geo. So it could be a multi step comparison, maybe saving a recent could be made async.

But I agree that we can get back to these questions, improvements later in separate stories.

I think querying the full history on typing would be really useful, so I would be happy to see a story for that.

About purging, just an idea that it could be done it `Remembering`, iterating over `Prioritized` and `forget()` elements after N count, maybe done in a worker thread.

lemonboston:
I think actually the equality check and purging can be both done at the point of saving selected item, so in `Remembering`. If either id() or name() and extraContext() is same as any `Recent` stored already, that one is forgotten.

I tried this, worked for the language duplicate:

```
new Thread(new Runnable()
        {
            @Override
            public void run()
            {
                SharedPreferencesRecents<GeoPlace> recents = new SharedPreferencesRecents<>(mAppContext, "places", new GeoPlaceCharSequenceConverter());

                Prioritized<GeoPlace> orderedRecents = new Prioritized<>(recents);

                int count = 0;
                for (Recent<GeoPlace> recent : orderedRecents)
                {
                    NamedPlace namedPlace = recent.value().namedPlace();
                    NamedPlace selectedNamedPlace = selectedGeoPlace.namedPlace();
                    if (namedPlace.id().equals(selectedNamedPlace.id())
                            ||
                            (namedPlace.name().equals(selectedNamedPlace.name()) &&
                                    namedPlace.extraContext().equals(selectedNamedPlace.extraContext())))
                    {
                        recent.forget();
                    }

                    if (count++ > 100)
                    {
                        recent.forget();
                    }
                }
                recents.recent(selectedGeoPlace).remember();

            }
        }).start();
```
The comparison would be better put in an `Equalable` decorator of course.
This way the equality of `Recents` is managed outside of it. It doesn't have to know about that. Not sure if it's better or worse, but seems good enough for now.

dmfs:
@lemonboston please check again. I've addressed some of your remarks with the latest commit.
Instead of moving the "recents" string (which should be "locations", btw) to a constant that's references twice I've created a class that hide the actual implementation details. I think that's a better approach.

Also, I've created a decorator for the module factories to store the recent locations. I agree that this is the better approach and allows to determine which modules add to the recent locations and which don't.

FYI: I've chosen to store the GeoPlaces as a URI, because it's a standardized representation for geo locations. Admittedly, the query parameters are not standardized, but I think that's not much of an issue. On top of that it's an easy way to serialize and parse geo locations/places.

dmfs:
I didn't address purging recent locations yet. IMO that's not an urgent issue. Even if you have 1000 recent locations the amount of data is still relatively small.

lemonboston:
I like the solution for the “locations” literal with `RecentGeoPlaces`.

COARSE_LOCATION can be omitted in demo manifest as FINE_LOCATION contains it according to:
http://stackoverflow.com/a/15310805/4247460

`RecentLocationsModule` 
At line 97, it can be simplified as:
```
mUpdateListener.onUpdate(new Clear<ListItem>(newQuery));
```
Line 121 to: (using `ReplaceAll` from `resultupdates` package)
```
mUpdateListener.onUpdate(new ReplaceAll<>(places, newQuery));
```

`if (count++ > 5)` - could the number of recents shown be a constant or constructor param?

`Remembered`
I actually meant applying the `ItemChosenAction` decorator within the module factories, not a factory decorator, but this works as well, not sure which one is better. Although I started to think that it might be easier to follow if we try to avoid decorating the factories, just have one factory for a module, let’s say `CurrentLocationModuleFactory` and it applies all configuration (ie decorators, proxy) that it needs at a single place. Not that important anyway.
I would still make that `Remebered(ItemChosenAction)` class and use it in `Remebered(SearchModuleFactory)`, even maybe just to not carry the object captured by the anonymous class further.

About purging and solving duplicates, could you please consider applying that code snipped above, in `Remebered`?
I paste it here again in full:
(Would be better in `Remembered(ItemChosenAction)` and the sameness check could be two `Equalable` decorator.)
```
@Override
    public SearchModule create(final Activity activity, ResultUpdateListener<ListItem> updateListener, final ItemChosenAction<GeoPlace> itemSelectionAction)
    {
        return mDelegate.create(activity, updateListener, new ItemChosenAction<GeoPlace>()
        {
            @Override
            public void onItemChosen(final GeoPlace selectedGeoPlace)
            {
    itemSelectionAction.onItemChosen(selectedGeoPlace);

                new Thread(new Runnable()
                {
                    @Override
                    public void run()
                    {
                        Recents<GeoPlace> recents = new Prioritized<>(new RecentGeoPlaces(activity));

                        int count = 0;
                        for (Recent<GeoPlace> recent : recents)
                        {
                            if (isSamePlace(recent.value().namedPlace(), selectedGeoPlace.namedPlace()))
                            {
                                recent.forget();
                            }

                            if (count++ > 100) // TODO move to constant
                            {
                                recent.forget();
                            }
                        }
                        recents.recent(selectedGeoPlace).remember();

                    }
                }).start();
            }
        });
    }

private boolean isSamePlace(NamedPlace namedPlace, NamedPlace selectedNamedPlace)
    {
        return namedPlace.id().equals(selectedNamedPlace.id())
                ||
                (namedPlace.name().equals(selectedNamedPlace.name()) &&
                        namedPlace.extraContext().equals(selectedNamedPlace.extraContext()));
    }

```
This solves possible duplicate questions (language duplicate, or same name but different location) and also the purging without any further change.

`SerializableCharSequenceConverter` - not used, is there any plan for it?

lemonboston:
The suggested code snipped could also be done as part of this: https://github.com/schedjoules/android-event-discovery-sdk/issues/145
So no need to spend time with it now.

dmfs:
I've changed update operations as mentioned.

As mentioned before, I don't want to address purging and duplicates with this PR, mostly because it's not an issue at this time. When we address it, it will certainly handled by the `Recents` implementation or a decorator of it, rather than a procedure like this. E.g. `Limited` could be a decorator to `Recents` that "forgets" all `Recent` objects after a specific number of objects have been iterated. So no need for an external procedure. This also goes for the duplicates. That's just a matter of how the values are stored. E.g. if we use the `NamedPlace` as the key, deduplication will happen automatically.

`SerializableCharSequenceConverter` was just my first implementation of this interface. Mostly to test the concept as such. I think it's ok to keep it.

The `Remembered` decoration should be done at the highest possible level. Since there is no harm in decorating the factory I prefer to keep it like this.

lemonboston:
Ok, review is done from my side, I suppose this is ready for rebase to master after #95 is merged, and then for merging.
(The 2 notes haven't been addressed: coarse permission in manifest, and number of items to be shown as contant, none of which is important though).

PR #149, opened 2017-03-15
Ad hoc/143 coverage

dmfs:
@lemonboston please note that this is also based on #95


lemonboston (commented):
`ACCESS_FINE_LOCATION` in demo manifest:
Can we add some comment there? Partners probably look at the demo app as a blueprint for integration, so just as there is a comment for how and when to remove the `uses-permission-sdk-23`, we could comment this one too. Something like “Integration doesn’t require ACCESS_FINE_LOCATION to be declared in your manifest, it is just for this demo app. SDK declares it with `uses-permission-sdk-23`, so it will be asked runtime only at the place of usage. If you have targetSdkLevel<23 though, remove it by adding the below to your manifest, to avoid requesting permission on installation.”

Coverage check
I think it would be good to provide a synchronous way as well to check the coverage. Partners might have their own way already to do async calls, eg with `AsyncTask`s or `RxJava`, they can handle that part, so I wouldn’t force them to use our asycn callback solution (pigeons), it may add overhead for them to understand and may not fit well in their approach.

`EventListFragment`
I tried and `SerializableDovecote<Serializable>` can be `SerializableDovecote<Boolean>` to be more type safe and avoid casting.
`Snackbar.make(getView()...` can be changed to `Snackbar.make(mViews.getRoot()..` to avoid the warning.

`CountryCode`
`getNetworkCountryIso()` javadoc says to check for `getPhoneType()`, so shouldn’t that be used, instead of `tm.getNetworkType() != TelephonyManager.NETWORK_TYPE_CDMA` ?

`CountryCode` `SimCountryCode` `toString()` `synchronized (this)`
Why is synchronization here? I think the certain synchronization cost exceeds the benefit of the extremely rare case of concurrently accessing the same cache. As I see there is no need for guaranteeing returning the same value instance here.

`LastKnownLocation`
Before iterating over the providers, there could a try with the one that has the best accuracy to maybe avoid using not the most accurate location:
```
Criteria criteria = new Criteria();
criteria.setAccuracy(Criteria.ACCURACY_FINE);
String bestProvider = locationManager.getBestProvider(criteria, true);
//noinspection MissingPermission
Location location = locationManager.getLastKnownLocation(bestProvider);
```

I changed the language to Dutch on my phone and still saw the not supported message, is that okay?

dmfs:
> |ACCESS_FINE_LOCATION| in demo manifest:
> Can we add some comment there? Partners probably look at the demo app 
> as a blueprint for integration, so just as there is a comment for how 
> and when to remove the |uses-permission-sdk-23|, we could comment this 
> one too. Something like “Integration doesn’t require 
> ACCESS_FINE_LOCATION to be declared in your manifest, it is just for 
> this demo app. SDK declares it with |uses-permission-sdk-23|, so it 
> will be asked runtime only at the place of usage. If you have 
> targetSdkLevel<23 though, remove it by adding the below to your 
> manifest, to avoid requesting permission on installation.”
>
Good point. I'll add somthing like that.
>
> Coverage check
> I think it would be good to provide a synchronous way as well to check 
> the coverage. Partners might have their own way already to do async 
> calls, eg with |AsyncTask|s or |RxJava|, they can handle that part, so 
> I wouldn’t force them to use our asycn callback solution (pigeons), it 
> may add overhead for them to understand and may not fit well in their 
> approach.
>
If partners want to do that synchronously they can directly perform the 
`Coverage` call (i.e. what the Service does). It's not that difficult. 
Later we might offer a `ContentProvider` that can be queried.
>
> |EventListFragment|
> I tried and |SerializableDovecote<Serializable>| can be 
> |SerializableDovecote<Boolean>| to be more type safe and avoid casting.
> |Snackbar.make(getView()...| can be changed to 
> |Snackbar.make(mViews.getRoot()..| to avoid the warning.
>
Ah, true. I'll change that.
>
> |CountryCode|
> |getNetworkCountryIso()| javadoc says to check for |getPhoneType()|, 
> so shouldn’t that be used, instead of |tm.getNetworkType() != 
> TelephonyManager.NETWORK_TYPE_CDMA| ?
>
I'll look into that again. I also planned to make them implement 
`Optional<CharSequence>` instead of `CharSequence`, so we don't have to 
fall back to the region.
>
> |CountryCode| |SimCountryCode| |toString()| |synchronized (this)|
> Why is synchronization here? I think the certain synchronization cost 
> exceeds the benefit of the extremely rare case of concurrently 
> accessing the same cache. As I see there is no need for guaranteeing 
> returning the same value instance here.
>
It's just copy&pasted from `CountryCode`, I didn't care about the 
details. But I guess I can remove it.
>
> |LastKnownLocation|
> Before iterating over the providers, there could a try with the one 
> that has the best accuracy to maybe avoid using not the most accurate 
> location:
>
> |Criteria criteria = new Criteria(); 
> criteria.setAccuracy(Criteria.ACCURACY_FINE); String bestProvider = 
> locationManager.getBestProvider(criteria, true); //noinspection 
> MissingPermission Location location = 
> locationManager.getLastKnownLocation(bestProvider);|
|I think the accuracy is less important here. It would be more 
interesting to get the most recent last known location. |
> ||
>
> I changed the language to Dutch on my phone and still saw the not 
> supported message, is that okay?
>
Currently, geo position and operator country have higher priority than 
the region. Try again with mobile data and GPS turned off.

lemonboston:
Checked the update:

I like that too that `NetworkCountyCode`, `SimCountryCode` and `LastKnownLocation` become `Optional<>` looks clearer this way.

You can use `AndroidGeoLocation` in `LastKnownLocation` instead of `StructuredGeoLocation`.

There is no localization for the new string resource yet.

Otherwise, looks good to me.

dmfs:
Since users of our partner apps should not see the error message, I'll merge this now. Translation of the error will be subject to another story.

Issue #156, opened 2017-03-17
Crash when selecting a result from the suggestions


dmfs:
```
03-17 21:09:05.574  6660  6660 E AndroidRuntime: Process: com.schedjoules.eventdiscovery.demo, PID: 6660
03-17 21:09:05.574  6660  6660 E AndroidRuntime: java.lang.IllegalStateException: Adapter is not attached to RecyclerView. Enable sticky headers after setting adapter to RecyclerView.
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at eu.davidea.flexibleadapter.helpers.StickyHeaderHelper.attachToRecyclerView(StickyHeaderHelper.java:78)
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at eu.davidea.flexibleadapter.FlexibleAdapter$5.run(FlexibleAdapter.java:1395)
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:815)
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:104)
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:207)
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:5688)
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:888)
03-17 21:09:05.574  6660  6660 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:749)
```

lemonboston:
Status:
I couldn't reproduced it at first, but thought from the error message that simply the `FlexibleAdapter` initialization order has to be changed. But after changing that and testing, the crash happened when I rotated the location picker screen and then selected a city. So the real problem is execution order when `EventListFragment` is standing up after both rotation and `onActivityResult`. So I am working on fixing that, some of that also relates to state management in this retained fragment.

dmfs:
So once again the problem is caused by retained fragments? Can't we just get rid of it?

[...]

PR #159, opened 2017-03-20
PR for 156: Fix crash when rotated location picker item is selected

lemonboston:
@dmfs Please review.

Besides the comment here: https://github.com/schedjoules/android-event-discovery-sdk/issues/156#issuecomment-287857275 , there is one more interesting change in the PR:
I've started to use `Initialization` interface, similarly as I mentioned in an email a few months ago. Looks like this (the earlier proposal had no method parameter):
```
public interface Initialization<O>
{
     void init(O operand);
}
```
I believe it could be used at several places later for initialization code usually in Activity/Fragment, to extract them if they form a small cohesive configuration and also to be able to re-use them.
I've added some related types and classes as well that are not used now but could be useful later, but I can also remove them if that's better.
There could be more than one 'operand' for an initialization, for example setting up the toolbar title operates on a `Toolbar` and the `Actvity`, so that could be an example for the 2 param version.
(Like `new SetToolbarTitle("Hi, there").init(toolbar, activity);`)
Please check them and slack me if you have concern over this general approach, I am curious what you think.


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/initialization/Initialization.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * General interface for representing any kind of initialization for a type of mutable object.
+ *
+ * @param <O>
+ *         type of the object to initialize
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface Initialization<O>
```
dmfs: This interface looks really wrong. This essentially resembles procedural programming in form of a class.

This
```java
        FlexibleAdapter<IFlexible> adapter = new FlexibleAdapterFactory().create();
        if (!shouldClear && mAdapter != null)
        {
            new CopyItemsFromOld(mAdapter).init(adapter);
        }
```
could be refactored with a FlexibleAdapterFactory decorator into something like this
```java
        Factory<FlexibleAdapter<IFlexible>> adapterFactory = new FlexibleAdapterFactory();
        if (!shouldClear && mAdapter != null)
        {
            adapterFactory = new Copying(adapterFactory, mAdapter)
        }
        FlexibleAdapter<IFlexible> adapter = adapterFactory.create();
```
with
```java
public final class Copying extends Factory<FlexibleAdapter<IFlexible>>
{
    public FlexibleAdapter<IFlexible> create()
    {
        // pass on create() and copy the data from the given adapter
    }
}
```

dmfs (changes requested):
Don't introduce procedural patterns through the back door ;-).
The interface name "Initialization" should make you suspicious.

lemonboston:
I think we haven't found a good approach yet for the initialization, configuration, setup code which usually resides in Activities, Fragments, I suppose because those are more procedural in nature by default.
You liked this proposal a few months ago ;)
So I see that we basically encompass a initialization function into an object here, but I also see it more fitting to our representative, declarative approach. How is different from  `ListChange.apply(list, adapter)`? I think they are similar, I even thought about the verb `apply` for the method name here. So we encapsulate some kind of change/initialization/setup/configuration for a type of mutable object, we declare it, represent it and then we can apply it later to the object. This allows for reuse, decorators, etc.

The factory decorator you suggested can work here, but in general this initialization approach is mostly aimed at objects that we don't instantiate ourselves. Views, activity configuration, etc. So most of the code that is still dumped into Fragment/Activities.

If you still not at ease with this, I remove it of course, let me know.

lemonboston:
As discussed in email, I've removed the `Initialization` interface and related types and used a factory decorator for `FlexibleAdapter` item copying.
Squashed and pushed, please review.

dmfs (approved):

Issue #162, opened 2017-03-22
Create fix layout for details page elements instead of the dynamically populated actions


lemonboston:
## In order to use the app effectively, as a user I want all UI elements on the details page to have a fix place where I can expect them.

In order to switch to this approach, we need to have a fix layout where the items have a fix placeholder.

lemonboston:
@rutgerg , @dmfs , @wilfredstegeman 
Just to be on the safe side, could someone please confirm that the API only sends 'add to calendar', 'share', 'directions' actions currently, so we don't already use any of the earlier defined actions?:

info, call, buy, remind, watch, invite

(They are supported now in the app, but if we switch to fix layout we need to define place for them, if they are used.)

dmfs:
This is not about what the API sends, but what we're going to present to the user. For now that's "add to calendar", "share" and "directions".

lemonboston:
I think this is not worth starting until #161 , #163 and #164 are reviewed and merged.

lemonboston:
@dmfs , @wilfredstegeman Could you please provide translation for the new share button label?:
Tip a Friend
(It will be displayed in caps.)

wilfredstegeman:
@lemonboston Keep 'Share' for now. Not sure about the Tip a friend at this moment. In Dutch that is 'Delen'.

PR #163, opened 2017-03-23
PR for 136: Ticket button

lemonboston:
@dmfs Please review.

I changed the accent color to the green from the design.

Small technical note: The ticket button view could be handled similarly as you suggested for the acttion views:
`new SimpleSmartActionView(view).update(new OptionalAction(actions, ADD_TO_CALENDAR))` 
But I would do that as part of #162 with all the other views.


lemonboston:
Added the translations and adjusted ticket button width.

eventdiscovery-sdk/src/main/res/values/strings.xml
```diff
@@ -26,6 +26,9 @@
     <string name="schedjoules_action_book_hotel">Hotel</string>
     <string name="schedjoules_action_book_parking">Parking</string>
     <string name="schedjoules_action_add_to_calendar">Add to calendar</string>
+
+    <string name="schedjoules_ticket_button">Get Tickets</string>
```
dmfs: Don't forget the describing comment. Also please add the context to the name, i.e. it should read like `schedjoules_event_details_ticket_button_title`. In future we may have other ticket buttons on other pages.

Note, the comment is only required in the default language.

eventdiscovery-sdk/src/main/res/layout/schedjoules_fragment_event_details.xml
```diff
@@ -56,9 +56,12 @@
                         android:orientation="vertical"
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
-                        android:layout_marginBottom="8dp"
                         android:layout_marginTop="12dp"/>
 
+                <android.support.v4.widget.Space
```
dmfs, line 61: This should be hidden by default and only become visible when the ticket button becomes visible too. Otherwise we have to much white space underneath events without ticket option.

dmfs (changes requested):

eventdiscovery-sdk/src/main/res/values/strings.xml
lemonboston: Added the missed comment and changed the name.

eventdiscovery-sdk/src/main/res/layout/schedjoules_fragment_event_details.xml
lemonboston: Oh, right, good catch, thanks. Updated.

lemonboston:
This is ready for review again.

dmfs (approved):

PR #164, opened 2017-03-24
PR for 140: Expandable description

lemonboston:
@dmfs Please review.

Note: I plan to refactor the Views used on the details screen, make them SmartView if possible along with the layout refactor as part of #162. So here I copied and followed the existing way for the expandable view.


dmfs (changes requested):
HTML links don't work anymore.

lemonboston:
Oh, that was bad. I added trimming whitespaces from the end of the parsed description as a last step and that broke the html format it seems. I removed the trimming, it should work now.

I did that trimming because there was often a few extra lines at the end of the content, but we probably need a better way to remove that if we wanted to.

Ready for review again.

dmfs (approved):

PR #172, opened 2017-03-30
PR for 174: Fix zombie GoogleApiClients

lemonboston:
Small changes for GoogleApis related code.

Ready for review,
@dmfs Could you please maybe test on real device that this line of code is actually executed when there is a resolvable Play Services related error?:

```
com/schedjoules/eventdiscovery/framework/googleapis/requests/ConnectionErrorHandling.java:60

if (connectionResult.hasResolution())
{
     // This is being auto-managed
     throw new GoogleApiAutoManagedIssueException(connectionResult.toString());
}
```
On the emulator I have, `hasResulution()` is always `false` even if it shows a dialog to update Play Services. It may be because eventually it can't be updated, when I press the button, nothing happens. It might different on real device.

Regarding the removed synchnorization for the `Set` in `BasicGoogleApis`, I know it means an `Api` could be readded and the client rebuild in a rare case, but I think that's okay, doesn't cause error, so I trade that with having to lock on the `Set` for every call.
But the main point of the change in the class there is to `disconnect()` the old `GoogleApiClient`.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/googleapis/BasicGoogleApis.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/googleapis/BasicGoogleApis.java
@@ -18,16 +18,16 @@
 package com.schedjoules.eventdiscovery.framework.googleapis;
 
 import android.support.v4.app.FragmentActivity;
+import android.util.Log;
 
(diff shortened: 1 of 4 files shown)


lemonboston:
I've created a bug for the zombie `GoogleApiClient` and re-linked this PR to that. (The branch name stayed the same.)

PR #173, opened 2017-04-03
PR for 166: Fix Current Location 'fllickering'

lemonboston:
The solution applied is to simply fire a caching call for current location once the screen loads. When user taps the call is done again, but it will likely be served quickly from the cache.
The "Find location.." progress item has been removed.

In case the call would take long and the user would press "Current location" quickly, before the caching call finishes, there can be some time between user tapping and screen disappearing, there is no progress indicator for that now. If that's important at this point, that could be a separate ticket I think. I have some idea about `ListItem` updates for that.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/CurrentLocationModule.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/CurrentLocationModule.java
@@ -74,7 +74,6 @@ public final class CurrentLocationModule implements SearchModule
     private final Lazy<Geocoder> mGeocoderFactory;
     private final GoogleApis mGoogleApis;
     private final ExecutorService mExecutorService;
-    private final GetCityTaskCallback mGetCityTaskCallback;
     private final Cache<ParcelableGeoLocation, GeoPlace> mCurrentCityCache;
 
 
@@ -92,7 +91,6 @@ public final class CurrentLocationModule implements SearchModule
 
         mExecutorService = Executors.newSingleThreadExecutor();
 
-        mGetCityTaskCallback = new GetCityTaskCallback();
         mCurrentCityCache = new TimedSingleValueCache<>(CACHE_CITY_MINUTES, TimeUnit.MINUTES);
     }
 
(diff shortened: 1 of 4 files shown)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/location/CurrentLocationModule.java
```diff
@@ -107,22 +105,54 @@ public void shutDown()
...
+            cacheCurrentLocation();
+
+            ListItem item = new Clickable<>(
+                    new MessageItem(mActivity.getString(R.string.schedjoules_location_picker_current_location)),
+                    new OnClickAction()
+                    {
+                        @Override
+                        public void onClick()
+                        {
+                            loadCurrentLocation();
```
dmfs, line 119: Why do you load the current location on a click and not in advance?

lemonboston: A few lines above a load is started (`cacheCurrentLocation()`). It's not the perfectly optimized solution, but it looked good enough now with some advantages, namely that if screen stays on and user moves, and click, it will use the latest location. But if it is the same as when the screen loaded, it will return quickly.
So there are 2 calls involed, the first for `getLastKnownLocation()` which is actually an instant synchronous call, just in our design now it is async, but still very fast. That is always executed, not cached. The second call for `GeoCoder` to get the city (address) is the slower, that is cached for a `GeoLocation`.
This way also when user clears the query, the caching call starts again.
The caching call has no callbacks, so if there is any error there, or simply user presses in the meantime, it won't cause any problem.
As mentioned in the description, there is no progress indication currently, which is only needed if user presses right away or cache call failed maybe. So not a frequent case. I would do a improvement story for that after #175.

dmfs:
I was wondering why you don't call `loadCurrentLocation` when the call in `cacheCurrentLocation` finishes? At present it looks like the user has to wait for `loadCurrentLocation` to finish when he clicked the button (what happens if he clicks the button multiple times before the request finishes?). Can't we start this right away, so the result is already available (at a good chance) when the user clicks the button?

lemonboston:
> Can't we start this right away, so the result is already available (at a good chance) when the user clicks the button?

That is what happens now, maybe the names of the methods are not good, these could be the full names:
`loadAndCacheCityForCurrentLocation()`
`loadAndSelectCityForCurrentLocation()`
So the the second method call also ends up selecting that city, closing the screen, we can only call that when user pressed. And this second call of course uses the same cache, so if the 1st one finished, it uses its result.
The only missed opportunity is when the call is let's say 3 sec and user pressed after 2 sec, then instead of 1 sec, they will wait 3 sec, but I am not sure it would be worth optimizing for this rare case, saving that state can increase complexity with keeping the renewed last location cases in mind as well.

> what happens if he clicks the button multiple times before the request finishes?

I tested that, it doesn't seem to cause any trouble. After the first finishing result, it is selected, Activity finished. The executor is serial and shutdown in onDestroy(), also the the `SafeAsyncTasks` involved are safe for nulled out callbacks. Also, I just looked into `Activity.setResult()`, it's only setting a field, so even if a second callback result would get through after let's say onDestroy(), it wouldn't cause problem.

dmfs (approved):

Issue #168, opened 2017-03-28
action text capitals


dmfs:
In order to identify clickable actions quickly, as user I want actions to be styled like buttons.
---
Actions should use a borderless button style and span the entire line (i.e. don't use a `wrap_content` width) The only concession we make to the design is that actions have black text (`#515457` to be more precise).

lemonboston:
I suppose it's worth doing after #162 or as part of that.

lemonboston:
Closing this, as this is addressed in #178 and #183 .

PR #178, opened 2017-04-05
PR for 162: Refactor event details layout for new fix elements

lemonboston:
@dmfs Could you please have a first look on how the `SmartView<Optional<T>>` classes are used now in `EventDetailFragment` along with the layout `include`s and bindings? It looks really useful and there are possibilities to create some more reusable elements there but I don't want to overdo it before you have look.

The task is not completely finished yet, the code part is mostly done, but I need to go through all the looks, colors, sizes, etc, so I will keep commiting to the branch.


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/ActionExecutingOnClickListener.java
```diff
@@ -0,0 +1,45 @@
...
+import android.app.Activity;
+import android.view.View;
+
+
+/**
+ * A {@link View.OnClickListener} that executes the given {@link ActionExecutable} on onClick().
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class ActionExecutingOnClickListener implements View.OnClickListener
```
dmfs: Maybe call it just `ActionClickListener` everything else seems kind of redundant.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/DirectionsLabelFactory.java
```diff
@@ -44,6 +44,6 @@ public DirectionsLabelFactory(Iterable<Location> locations)
     @Override
     public CharSequence create(Context context)
     {
-        return new AddressName(mLocations).value(context.getString(R.string.schedjoules_action_directions));
+        return new AddressName(mLocations).value(context.getString(R.string.schedjoules_action_directions).toUpperCase());
```
dmfs, line 47: Isn't the directions label styled like a BorderlessButton? If possible leave the up-case conversion to the styles.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/OptionalAction.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * {@link Optional<Action>} that looks for the {@link Action} in a provided {@link Iterable<Action>} by link rel type.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class OptionalAction extends AbstractCachingOptional<Action>
+{
+
+    public OptionalAction(final String actionLinkRelType, final Iterable<Link> actionLinks, final ActionFactory actionFactory, final Event event)
```
dmfs, line 41: Why do you pass an `ActionFactory` to this constructor?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/OptionalAction.java
```diff
@@ -0,0 +1,64 @@
...
+                Iterator<Link> filteredIterator = new FilteredIterator<>(actionLinks.iterator(), new AbstractFilteredIterator.IteratorFilter<Link>()
+                {
+                    @Override
+                    public boolean iterate(Link link)
+                    {
+                        return link.relationTypes().contains(actionLinkRelType);
+                    }
+                });
+                //noinspection ConstantConditions
+                return filteredIterator.hasNext() ?
```
dmfs: Maybe create a `First` adapter, just like the one in the Optional package. The one in the Optional package is only for `Iterables`, but in this case it would make sense to have one for Iterators too. The downside is that you would have to execute the check eagerly in the constructor (because the Iterator is not immutable).

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/SimpleAction.java
```diff
@@ -50,7 +51,7 @@ public SimpleAction(CharSequenceFactory labelFactory, @DrawableRes int iconResId
 
     public SimpleAction(@StringRes int labelResId, @DrawableRes int iconResId, @NonNull ActionExecutable actionExecutable)
     {
-        this(new ResourceCharSequenceFactory(labelResId), iconResId, actionExecutable);
+        this(new UpperCased(new ResourceCharSequenceFactory(labelResId)), iconResId, actionExecutable);
```
dmfs, line 54: Again, the upper case conversion should probably be handled by the style.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/TicketButtonAction.java
```diff
@@ -0,0 +1,96 @@
...
+{
+
+    public TicketButtonAction(final Optional<Link> ticketLink, final Event event)
+    {
+        super(new Factory<Optional<Action>>()
+        {
+            @Override
+            public Optional<Action> create()
+            {
+                if (!ticketLink.isPresent())
```
dmfs: I wonder if we can achieve the purpose of this method with a decorator. Something like a "MappedOptional" (similar to the Converted Iterator decorator).
like so:
```java
return new Mapped<Action>(ticketLink, new Function<Link, Action>(){
      public Action mapped(Link link)
      {
        return new InternalTicketAction(link, event);
      }
   });
```

Not sure if we would still need TicketButtonAction in that case.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/ActionView.java
```diff
@@ -0,0 +1,71 @@
...
+        {
+            final Action action = optAction.value();
+            mTitleView.setText(action.label(mTitleView.getContext()));
+            mIcon.setImageDrawable(action.icon(mIcon.getContext()));
+            mRoot.setOnClickListener(new ActionExecutingOnClickListener(action.actionExecutable()));
+            mRoot.setVisibility(View.VISIBLE);
+        }
+        else
+        {
+            mTitleView.setText(null);
```
dmfs: Is it really necessary to set null values? Isn't is sufficient to set visibility to GONE?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/TicketButtonView.java
```diff
@@ -0,0 +1,65 @@
...
+        mButtonArea = views.schedjoulesEventDetailsTicketButton.getRoot();
+        mButton = views.schedjoulesEventDetailsTicketButton.schedjoulesEventDetailsTicketButton;
+        mSpace = views.schedjoulesEventDetailsTicketButtonSpace;
+    }
+
+
+    @Override
+    public void update(Optional<Action> bookAction)
+    {
+        mButtonArea.setVisibility(bookAction.isPresent() ? View.VISIBLE : View.GONE);
```
dmfs: better just call `bookAction.isPresent()` just once and branch with an `if` statement. It might become a bit longer but it's more readable and cleaner.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/OptionalTextedView.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class OptionalTextedView implements SmartView<Optional<CharSequence>>
+{
+    private final OptionalView mRoot;
+    private final TextView mTextView;
+
+
+    public OptionalTextedView(View root, TextView textView)
```
dmfs: why does it need both, the View and the Root? Could you pass the view resource id instead of the view itself?

eventdiscovery-sdk/src/main/res/layout/schedjoules_fragment_event_details.xml
```diff
@@ -24,14 +24,78 @@
...
+                        android:id="@+id/schedjoules_event_details_description"
+                        android:layout_width="match_parent"
+                        android:layout_height="wrap_content"
+                        android:layout_marginLeft="16dp"
+                        android:layout_marginRight="16dp"
+                        android:layout_marginTop="8dp"
+                        android:layout_marginBottom="8dp"/>
+
+                <include
+                        layout="@layout/schedjoules_view_event_details_action"
```
dmfs: a lot of repetition here. Please create a separate style. Also watch out for the box model. Often it's better to give the element the same margin on all sides (e.g. 8dp) and have the missing 8dp added by a padding on the container element.

eventdiscovery-sdk/src/main/res/values/dimens.xml
```diff
@@ -13,7 +13,7 @@
     <dimen name="schedjoules_list_item_image_width">64dp</dimen>
     <dimen name="schedjoules_list_item_default_height">96dp</dimen>
     <dimen name="schedjoules_event_list_progress_bar">36dp</dimen>
-    <dimen name="schedjoules_event_detail_item_icon">28dp</dimen>
+    <dimen name="schedjoules_event_detail_item_icon">24dp</dimen>
```
dmfs: What's that? Height? Width? Margin? Padding? Spacing? Elevation? … Please add what kind of value this is.

dmfs (changes requested):
Done with the first round.

dmfs:
Please also make sure the action items have a padding instead of a margin, otherwise the sensitive area becomes to small and the visual feedback covers only a small area too.

dmfs:
The footer doesn't become fully visible any more when you scroll down on a page with expanded description. Also see the attached image.

device-2017-04-06-000012

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/ActionExecutingOnClickListener.java
lemonboston: Done.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/DirectionsLabelFactory.java
lemonboston: The Directions action can have 2 labels now: DIRECTIONS in case we have no address and the address in normal case if we have it, although I am not sure if the first case would ever happen.
What would the BorderlessButton style add? Could we still simply add the icon there? Now the whole layout, the icon and the text acts as a button when pressed.
In any case, can we address it in a separate story if needed?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/OptionalAction.java
lemonboston: There was an `Actions` `Iterable` earlier that created the `Action`s, but I needed to filter them by rel type and instead of adding that as method to `Action` (which I did at one point), I thought it may be better to just filter the `Link`s first and then create that single `Action`. So `OptionalAction` needs the factory to create the `Action` from the corresponding `Link`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/TicketButtonAction.java
lemonboston: Created `Mapped` Optional for that and kept `TicketButtonAction` which now hides `BookTicketLink` as well.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/ActionView.java
lemonboston: No, it was not, right, I removed them. This still works correctly if it is updated multiple time.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/TicketButtonView.java
lemonboston: `OptionalView` is used now to hide that visibility switching.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/OptionalTextedView.java
lemonboston: This class has been removed.

eventdiscovery-sdk/src/main/res/layout/schedjoules_fragment_event_details.xml
lemonboston: Yeah, that was intentional at that point, I wanted to get back to styles and the end, also to see how the layout ends up. I've created a style for these event detail item sizes.
I left the 16dp horizontal padding instead of 8dp and then 8dp on the container, to have a greater area for the touch ripple effect.

eventdiscovery-sdk/src/main/res/values/dimens.xml
lemonboston: Changed to `schedjoules_event_detail_item_icon_size` as it is both height and width.

lemonboston:
Changed margins to paddings as suggested, also fixed the bug above, it was caused by the `LinearLayout` in the `NestedScrollView` having top margin, changing to padding fixed it.

lemonboston:
Ready for review.

As this changeset got big, I stop commiting to this and branch off for outstanding things that I have in mind:

- Venue name text size is not increased yet, because that causes misalignment of the icon - text. Need to find solution to that.
- Review the event details screen look again compared to design, check all sizes and colors
- Review colors.xml, styles.xml, make any common sense changes, then look through the event details related layouts again to use those or new styles where it make sense. Maybe also create named dimens for text sizes. Check if it make sense to apply primary and secondary text color theme properties. Maybe communicate color palette and usage to ui-ux channel. Ask about sizes on the design (there is a text with 23px for example.). Check divider color (the Android one looked too light).
- BorderlessButton for Actions if needed

I'll start with the first on a separate branch, we can create a bug for that.

lemonboston:
There is already a separate ticket for Action labels, we can address any possible change needed there:
https://github.com/schedjoules/android-event-discovery-sdk/issues/168

dmfs (approved):
Approving this despite concerns regarding the uppercase handling.
Also, make sure to submit a PR that restores the original name of schedjoules_fragment_event_details_action_loading

Issue #180, opened 2017-04-06
Upgrade http-client-essentials to 0.10.3


dmfs:
There is a potential NPE in 0.10.2 which has been fixed in 0.10.3

PR #182, opened 2017-04-06
Upgrade http-client-essentials version. Implements #180

dmfs:
@lemonboston please review with high priority


lemonboston (approved):
It looks good, builds for me.
I wonder thought whether the variable would be better to just being kept closer, declaring in the same file.

Issue #181, opened 2017-04-06
Update venue name size on details screen


lemonboston:

lemonboston:
This has been implemented on branch `stories/181-make-venue-name-bigger` but it is based on #178 , so when that's merged, I will update this and submit the pull request.

PR #184, opened 2017-04-07
PR for 181: Increase venue name size

lemonboston:
Implements #181.


dmfs (changes requested):
That structure looks way to complicated. A `RelativeLayout` nested in a `RelativeLayout` can't be right. Shouldn't a simple `drawableLeft` do the trick?

lemonboston:
From chat by @dmfs 

> regarding PR 184, when I reviewed that branch, I noticed that all icons we use are larger than they should be. (32dp vs. 24dp)
> I think the layout can be much easier if the icon size is correct, mostly because you probably can use "drawableLeft" instead of the rather complicated structure with an "align" element
> Please file an ad-hoc to replace the icons with correct sizes and to revisit the layouts

TODO after checking if drawableLeft would suffice create the stories as needed

lemonboston:
I've updated the layout, `drawableLeft` could work, indeed, but I still needed to keep a dummy, 1 line TextView for anchoring the icon to the first line.

I've created https://github.com/schedjoules/android-event-discovery-sdk/issues/191 to revisit the other icons and layouts, please move it to Sprint Backlog if you see fit.

Let me know if I should squash.

dmfs:
Looks better. Please squash.

lemonboston:
Squashed and rebased.

Issue #137, opened 2017-03-06
Refactor event details title


dmfs:
In order to see get the most important details of an event, as a user I want to see title and date (& time) on the tile.
---
There is a library that auto-scales textviews based on available space: https://github.com/AndroidDeveloperLB/AutoFitTextView

Also don't forget to update the loader microfragements with the new layout.

PR #185, opened 2017-04-07
Update event details header. Implements #137

dmfs:
* switch to custom multilinecollapsingtoolbarlayout (opacapp probably won't merge our changes)
* switch to a light appbar theme
* add Smart EventHeaderView to update the header
* adjust theme as needed

Note the UI design may not be final yet, but that will be subject to another story.

### DIFF
--- a/eventdiscovery-demo/src/main/res/values/strings.xml
+++ b/eventdiscovery-demo/src/main/res/values/strings.xml
@@ -1,6 +1,6 @@
 <resources>
-    <string name="app_name">SchedJoules Event Discovery Demo</string>
-    <string name="event_discover_now">Event Discovery\nfor today</string>
-    <string name="event_discover_later">Event Discovery\nin 10 days from now</string>
-    <string name="eventdetails_by_uid">Event Details by UID</string>
+    <string name="app_name" translatable="false">SchedJoules Event Discovery Demo</string>
+    <string name="event_discover_now" translatable="false">Event Discovery\nfor today</string>
+    <string name="event_discover_later" translatable="false">Event Discovery\nin 10 days from now</string>
+    <string name="eventdetails_by_uid" translatable="false">Event Details by UID</string>
 </resources>
--- a/eventdiscovery-sdk/build.gradle
+++ b/eventdiscovery-sdk/build.gradle
@@ -65,7 +65,7 @@ dependencies {
     compile 'com.github.dmfs:MicroFragments:e12ac1615e07fb0a9012f32267696a48fda0f69b'
     compile 'com.github.schedjoules:java-api-client:0.5'
     compile 'com.github.bumptech.glide:glide:3.7.0'
-    compile 'net.opacapp:multiline-collapsingtoolbar:1.3.0'
+    compile 'com.github.dmfs:multiline-collapsingtoolbar:79a3e4520261300cf32f2954832173bef2a43a40'
     compile 'com.google.android.gms:play-services-places:9.8.0'
     compile 'com.google.android.gms:play-services-location:9.8.0'
     compile('eu.davidea:flexible-adapter:5.0.0-rc1') {
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/datetime/DateTimeRange.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/datetime/DateTimeRange.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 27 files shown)


lemonboston:
The new header looks nice.

Some notes (I will continue checking the branch, please don't squash next commits.)


Changing location dropdown arrow to black is a small change I think, could be a separate bug as well of course.
`schedjoules_ic_arrow_drop_down_white.xml` to `schedjoules_ic_arrow_drop_down.xml`
`android:fillColor="#000000"` or maybe `android:fillColor="?attr/android:textColorPrimary"`

`DateTimeRange` - javadoc not updated with example of the new format

I’ve checked and 
```
if (mDateTime.getYear() == DateTime.nowAndHere().getYear())
        {
            // don't show year if it's this year
            format |= DateUtils.FORMAT_NO_YEAR;
        }

```
is not needed in `DateTimeRange` and `LongDateAndTimeNoYear`, that looks like to be the default behaviour of `DateUtils` formats.
Also, I just tried this
`DateUtils.formatDateRange(context, mStartTime.getTimestamp(), mStartTime.getTimestamp()..` and it works correctly, doesn’t show end time, so we may not need `LongDateAndTimeNoYear` at all, although that logic may be more appropriate then setting duration to zero.

`EventHeaderView`
`mHeader.schedjoulesEventDetailToolbarLayout.setExpandedTitleColor(0xffffffff);`
`upArrow.setColorFilter(0x80000000, PorterDuff.Mode.SRC_ATOP);`
I think we should start extracting these colors to at least a named color resources first, so we are able to better generalize them. For example the black and white text colors should probably come from the theme textColorPrimary attribute. But in order to get there to reference these correctly, I would suggest we name our colors as point of usage first. So we can at least see them in `colors.xml`.
`mActivity.getResources().getDrawable(R.drawable.abc_ic_ab_back_material);`
-> `ContextCompat.getDrawable(..)`

lemonboston:
Event list date headers are invisible now with the white background and white text.
I suppose we should not merge it without fixing that even if it is coming up in the next pull request, right?

`colors.xml`
`<color name="schedjoules_silver_chalice">#b0b0b0</color>`
`<color name="schedjoules_colorPrimaryDark">@color/schedjoules_silver_chalice</color>`
I see `#2E3135` with opacity 22% in InVision for the status bar color. Can we use that? Just so that we keep in sync with the design.

AppBar elevation doesn’t show up on API level 16, 17, 19 (the ones I checked), so there is no separator displayed there. 

(not finished with the review yet)

dmfs:
The date headers will be fixed by #188 (which is based on this branch). But I won't merge this until #188 has been approved to keep master "stable".

lemonboston:
Ok.

Last comments:

`EventDetailFragment`
Menu handling
It’s a bit hard to follow now with having both that `EventDetailsMenu` delegation and the other setup for the other toolbar. I think we could just remove `EventDetailsMenu` and do everything in the appropriate callbacks. Like moving both inflation to `onCreateOptionsMenu()`.
I had started a design earlier for dynamic, declarative Menu composition without using xml, so if we want to improve the code later, I think we could go into that direction. So that’s why I also think there is no need for that
`class EventDetailsMenu implements OnCreateOptionsMenuFragment, OnOptionsItemSelected`. 

`Toolbar toolbar = (Toolbar) views.getRoot().findViewById(R.id.schedjoules_event_detail_toolbar_light);`
can be
`Toolbar toolbar = views.schedjoulesDetailsHeader.schedjoulesEventDetailToolbarLight;`

lemonboston:
Just noticed that the even list circular ProgressBar is now white on white background.
Removing the tint here will change it to the green accent:
`layout/schedjoules_fragment_event_list.xml:62`

dmfs:
@lemonboston I'm done with the changes, please review again.

We'll create a separate ad-hoc to revisit the definition and usage of colors.

I prefer to be explicit with the year in the date-time strings and not rely on some undocumented platform feature.

lemonboston:
`DateTimeRange` javadoc update missed the year.

I see that this compat appbar shadow is a pain.
Just one obvious note to that, to perhaps extract the shadow View to a layout and include, since it seems to be the same everywhere.

`EventHeaderView`
This:
```
// set home icon of the light toolbar
final Drawable upArrow = mActivity.getResources().getDrawable(R.drawable.abc_ic_ab_back_material);
upArrow.setColorFilter(ContextCompat.getColor(mActivity, R.color.schedjoules_text_secondary), PorterDuff.Mode.SRC_ATOP);
mHeader.schedjoulesEventDetailToolbarLight.setNavigationIcon(upArrow);
```
can be changed to this:
```
// set home icon of the light toolbar
Drawable arrow = DrawableCompat.wrap(ContextCompat.getDrawable(mActivity, R.drawable.abc_ic_ab_back_material)).mutate();
DrawableCompat.setTint(arrow, ContextCompat.getColor(mActivity, R.color.schedjoules_text_secondary));
mHeader.schedjoulesEventDetailToolbarLight.setNavigationIcon(arrow);
```
(That `mutate()` is needed to not apply the change to that drawable resource globally)

lemonboston (approved):

dmfs:
thx. Merged it.

Issue #183, opened 2017-04-07
Fine tuning at the end of Update Design Phase 1


lemonboston:
We will collect here the small outstanding issues or discrepancies between the design and the look that were not worth addressing in the bigger stories.

This will also give us a chance to review the whole screen after all components are in place and compare to design again.

---

- [x] 1. Description text and icon alignments. Currently both have 16dp margin, but as the material icons have a standard 2dp padding, they are not aligned. Moreover some icons, like the _location pin_ one for the venue name is even slimmer, so has empty horizontal space beside the 2dp padding, so it's left edge won't align with the text, as it is done on the design.
Simply increasing description text margin to 18dp has the concern of leaving the 8dp grid.

- [x] 2. Directions address text size. It is 14sp size currently instead of 16sp as on the design, because it uses the same view as the other actions which have 14sp

- [x] 3. Action view text styles. The capital case is applied in code dynamically currently because the directions label have normal case. Directions could however get dedicated layout as well.
It was suggested to use style for capital case and maybe use BorderlessButton style. Get back to all this if needed. 

- [x] 4. Divider color. The 12% transparency one, which material specs say looked too light with the 1px height. Check again and decide.

- [x] 5. (Optional) Review colors, styles, dimens usage:
Review colors.xml, styles.xml, make any common sense changes, then look through the event details related layouts again to use those or new styles where it make sense. Maybe also create named dimens for text sizes. Maybe communicate color palette and usage to ui-ux channel.

- [x] 6. Check if it make sense to apply our primary and secondary text colors as theme properties. For example location picker item second line now use default theme color just as all TextViews where we may forget to specify, so I guess it would probably make sense to update the theme.
We may be able to use `tertiary` color as well, just found this:
```
        <!-- The most prominent text color.  -->
        <attr name="textColorPrimary" format="reference|color" />
        <!-- Secondary text color. -->
        <attr name="textColorSecondary" format="reference|color" />
        <!-- Tertiary text color. -->
        <attr name="textColorTertiary" format="reference|color" />
```
in `android-sdk/platforms/android-23/data/res/values/attrs.xml`. Although not sure what widgets use that and for what.

- [x] 7. Review distances between elements.
(One in particular: action icon and text distance is 18dp now, probably can be 24dp to fit the grid.)

- [x] 8. Event details loader screen doesn't look very nice (black title, loading message not in the middle) - see screenshot in commennts

lemonboston:
[...]

PR #188, opened 2017-04-10
Bugs/186 fix location picker theme

dmfs:
This updates the theme of the location picker. Note the branch is based on PR #185

### DIFF
--- a/eventdiscovery-sdk/src/main/AndroidManifest.xml
+++ b/eventdiscovery-sdk/src/main/AndroidManifest.xml
@@ -69,7 +69,7 @@
                 android:label=""
                 android:exported="false"
                 android:enabled="true"
-                android:windowSoftInputMode="adjustPan|stateAlwaysVisible"
+                android:windowSoftInputMode="adjustPan"
                 android:hardwareAccelerated="true"
                 android:theme="@style/SchedJoules_AppTheme.NoActionBar"/>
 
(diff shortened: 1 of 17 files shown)


lemonboston (changes requested):
It looks nice where these UI changes go. Location picker items look nicer to me, too, with the new text sizes.

Just curious, do you plan any divider, elevation to the bottom of the header items? Or does that still depend on what we come up for the event list date header?
The "Recent locations" feels a  bit detached or strange now for me.

`MessageItem` with its layout has been udpated for the new Current Location look, but it is used for unrecovarable errors as well. So I suppose we need another layout for that. I think it can be the same as the buttoned version, just without the button.

`schedjoules_list_item_place_suggestion.xml` - the extra content text has no color set now, so I suppose it uses the primary text color of the theme. It looks good imo, but the text color usage may start to be a bit inconsistent then. I updated a note about this primary, secondary text attribute in the fine tuning story, so we can address it there. https://github.com/schedjoules/android-event-discovery-sdk/issues/183

lemonboston:
It seems a bit too spacious vertically for me, I think smaller area could be enough to get comfortable tapping. That way we could maybe show more items (?).
(Not important of course, we can change later as well.)

`MessageItem` with its layout `schedjoules_list_item_location_message.xml` is still used for unrecoverable errors and for CURRENT LOCATION as well. Or was that deliberate?
If you check `MessageItem` constructor usage, you can see the other 3 use points.
So I think current location needs its own item and layout.

`<dimen name="schedjoules_list_subheader_height">48dp</dimen>` is not used.

dmfs:
The spacing follows material design for two-line list items. With the previous item height, it would be harder to distinguish items without having dividers. We could go for the dense version of the list item, but that would require a smaller font.

I'll check the subheader height dimension. Apparently i forgot to apply it :-o
I didn't check MessageItem, but I can look into that.

lemonboston (approved):

lemonboston:
All looks good to me now.

PR #193, opened 2017-04-11
update uri-toolkit fixes #179

### LINKED ISSUE #179: Upgrade uri-toolkit dependency to latest
This will also remove the old `Optional` type, so we can replace everything with the one from the dedicated optional library.

dmfs:


lemonboston (approved):
Looks correct to me.

PR #201, opened 2017-04-12
PR for 176: Remove empty lines from the bottom of event descriptions

lemonboston:


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/TrailingLinesTrimmed.java
```diff
@@ -0,0 +1,58 @@
...
+    // http://stackoverflow.com/a/25058538
+    // (Note: A Character.isWhitespace() might not be good for html, it messed up the html format before.)
+    private static CharSequence removeTrailingLines(CharSequence text)
+    {
+        if (text.length() == 0)
+        {
+            return text;
+        }
+
+        while (text.charAt(text.length() - 1) == '\n')
```
dmfs: This won't work if there is also spaces on these empty lines, also it's a very inefficient way of doing it.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/Html.java
```diff
@@ -15,31 +15,30 @@
...
-    public BasicStrings(Context context)
+    public Html(final String inputSource)
     {
-        mAppContext = context.getApplicationContext();
+        super(new Factory<CharSequence>()
+        {
+            @Override
+            public CharSequence create()
+            {
+                return new TrailingLinesTrimmed(android.text.Html.fromHtml(inputSource));
```
dmfs: This doesn't work as expected. It means all styles are dropped, because TrailingLinesTrimmed is not a Spannable anymore. `Html` has to implement the `Spannable` interface to make this work. To trim a Spannable, check this out: https://gist.github.com/mrmike/77fe2a6a357e78108115

dmfs (changes requested):

lemonboston:
I've reworked the trimmings and use a new `AbstractSpanned` base class for `Html`.

The plain text is trimmed from the leading lines with no content, but only from the full lines, so that identation on first line is not lost, and also trimmed from anything at the end. Its done with new `LeadingEmptyLinesAndTrailTrimming` `CharSequence` decorator.
I tried to use it for Html as well, and 'wrap' result in `Spanned` but styles was lost, I am not sure why. So I ended up using the gist you suggested, so for Html only empty lines are trimmed from both end.

Ready for review again.
(I squashed and rebased as I forgot not to, but almost all changes are reworked anyway.)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/Trimming.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
+
+
+/**
+ * Trimming {@link CharSequence} decorator.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class Trimming extends AbstractCharSequence
```
dmfs: The name should be "Trimmed", because it should describe the result not the operation.

Also, you don't seem to use this class and I don't seem much of a reason why we would trim spaces but not new lines. Consider to remove this and rename `LeadingEmptyLinesAndTrailTrimming` to `Trimmed`

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/EmptyLinesTrimming.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
+
+
+/**
+ * {@link Spanned} decorator that trims leading and trailing empty lines.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class EmptyLinesTrimming extends AbstractSpanned
```
dmfs: This should be `Trimmed` as well. There is no need to say that it trims empty lines.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/Linkified.java
```diff
@@ -0,0 +1,47 @@
...
+{
+    public Linkified(final CharSequence input)
+    {
+        super(new Factory<Spanned>()
+        {
+            @Override
+            public Spanned create()
+            {
+                SpannableString spannableString = new SpannableString(input);
+                Linkify.addLinks(spannableString, Linkify.ALL);
```
dmfs: Maybe consider to make `Linkify.ALL` a constructor parameter. Otherwise there is probably no way to customize this. How does this play with #210?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/LeadingEmptyLinesAndTrailTrimming.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
+
+
+/**
+ * {@link CharSequence} decorator that trims leading empty lines and trims the end of the text.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class LeadingEmptyLinesAndTrailTrimming extends AbstractCharSequence
```
dmfs: Why doesn't it trim spaces on the first line of text?
According to the unit test `"   \n  Hello"` becomes `"  Hello"`, I think it should become `"Hello"` without any leading white space at all.

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/Trimming.java
lemonboston: It was `Trimmed` first but I renamed to follow the decorator naming, like the `Following` or `UrlRewriting`. But `Trimmed` feels better to me, too, so I renamed it now.

It's not used, that's true, it just looked like something useful, so I kept it. Let me know if I should remove though.

About `LeadingEmptyLinesAndTrailTrimming` - see comment there.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/EmptyLinesTrimming.java
lemonboston: It only trims empty lines, not all whitespaces what I would expect from a general trimming. That's why the more specific name.
Should I still rename it?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/Linkified.java
lemonboston: Added the constructor parameter and a default constructor with `Linkify.ALL`.
Checked #210 problem with only `Linkify.WEB_URLS`, didn't help.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/LeadingEmptyLinesAndTrailTrimming.java
lemonboston: I thought it would be better to keep leading whitespaces on the first line, because it can be part of a formatting of the whole text. Like first-line indents for each paragraph. Or bullet points like " - text" where each line has a leading space. Or indents for section titles.

So that's why this ugly name of the class, it does different things on the leading and trailing end of the `CharSequence`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/EmptyLinesTrimming.java
dmfs: So why doesn't it trim all white spaces? I think that's exactly what it should do. I can't imagine why we would want a piece of HTML (that's what we're using it for) to start or end with any sort of white space.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/Trimming.java
dmfs: `Trimmed` represents a concrete value, i.e. the trimmed version of the decorated `CharSequence`. If we didn't have the rule to omit the base type for decorators the full name would be `TrimmedCharSequence`. It's a special version of the decorated `CharSequence` that's trimmed.

In contrast, `Following` and `UrlRewriting` modify the behavior of the decorated class.

If we had a factory that creates `CharSequences`, `Trimming` would be a decorator to that factory which trims the created `CharSequences`, i.e.
```
CharSequence s = new Trimming(new CharSequenceFactory()).create();
```

Does that make sense?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/LeadingEmptyLinesAndTrailTrimming.java
dmfs: In most cases I would want leading white spaces to be removed. Bullet points is really the only exception I can think of. However, a description that starts with a bullet point is hopefully very rare.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/EmptyLinesTrimming.java
lemonboston: This was taken from the snippet you posted and which I saw myself, too, somewhere, so just to have a well enough solution for this. I am a bit unsure about some parts of this spanned, charsequence trimming, so I didn't want to spend more time with this task to look into them more thoroughly.
Things like: which are the whitespace char codes, how best to check for them,
how does spanned work internally, is it possibly that it uses some non-visible chars for formatting purposes, can that cause problem if we strip it,
why is that that subsequence() didn't seem to work with spanned (lost styling) (this code snippet uses delete() ),
if we want to trim new lines, do we check for '\n' or '\r' as well.
Writing unit test for anything new also adds to the time.

So now that server-side might implement trimming https://github.com/schedjoules/event-discovery-backend/issues/45 , my question is, should we spend more time on this now or not?
We could use this version, and put a task to backlog to revisit it after server-side trimming has been implemented. At that time, I think we only need to trim trailing new lines for html spanned, nothing else, to remove those lines added by `Html` for paragraph endings.

lemonboston: Just realized that server won't trim the content of the html I suppose, so we need to do it on client side. So I will do this `Trimmed` `Spanned`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/Trimming.java
lemonboston: Yes, the distinction is clearer, thanks for elaboration.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/LeadingEmptyLinesAndTrailTrimming.java
lemonboston: Ok, I will use the `Trimmed` `CharSequence` then if you think that's better. And remove this one.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/EmptyLinesTrimming.java
dmfs: It doesn't, but it certainly should. As a client you shouldn't have to alter the content, just display it.
For the time being we should trim on the client though.

dmfs: Ok, we keep this as is for now. Let's see how much of an issue this actually is.

lemonboston: Ok, kept it, just renamed to `EmptyLinesTrimmed`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/charsequence/LeadingEmptyLinesAndTrailTrimming.java
lemonboston: Removed `LeadingEmptyLinesAndTrailTrimming` and now using `Trimmed` instead.

lemonboston:
Ready for review again.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/EmptyLinesTrimmed.java
```diff
@@ -0,0 +1,68 @@
...
+
+    // https://gist.github.com/mrmike/77fe2a6a357e78108115
+    private static SpannableStringBuilder trimSpannable(SpannableStringBuilder spannable)
+    {
+        int trimStart = 0;
+        int trimEnd = 0;
+
+        String text = spannable.toString();
+
+        while (text.length() > 0 && text.startsWith("\n"))
```
dmfs: Sorry, I missed the actual implementation before. It is rather inefficient as it creates a new String with every iteration. Consider using an implementation that works like in `Trimmed` (i.e. just count the number of chars that need to be removed instead of creating new strings without that char).

dmfs (changes requested):

dmfs:
please squash after these changes

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/spanned/EmptyLinesTrimmed.java
lemonboston: Updated with similar logic as in `String.trim()` (and in `Trimmed`) and created unit test.

lemonboston:
Squashed and rebased.

lemonboston:
Updated with a few more unit test cases (squashed, rebased).

dmfs (approved):

Issue #187, opened 2017-04-10
Align list view design with new white theme and card layout


dmfs:
In order to make the list view comply with the new white theme, we need to adjust the sections headers and replace the list items with cards.
---
Since the section headers are shared with the location picker, this part will be solved by #186

PR #203, opened 2017-04-12
update listview to work with white theme, implements #187

dmfs:


dmfs:
@lemonboston sorry, I had to merge, because I wanted this to be in today's release. Please review nevertheless and create bug ticket if you find any issues.

lemonboston:
Looks good to me.

Just one question:
`dimens.xml`:
Could we maybe join these 3 up in one resource?
```
schedjoules_activity_horizontal_margin
schedjoules_activity_vertical_margin
schedjoules_location_list_item_horizontal_margin
```
Something like `schedjoules_standard_margin` (or `..screen_margin`). All are 16dp, the 3rd one is used for not just location list but event list as well. So I think this 16dp really serves as a standard, default screen margin now. If we ever deviate from that somewhere, we would create the specific dimen then.
I know the first two are part of a new project template, but they are not used as theme attributes, so I think it’s okay to remove them.

dmfs:
At this time I would refrain from merging these. Once we start improving the tablet experience we'll certainly start using these values, because many margins have different values on larger devices. But that's a different story.
We probably should do to dimensions what I've done in #208 to colors.

Issue #192, opened 2017-04-11
Color tinted ProgressBar


lemonboston:
In order to reduce code duplication, improve code quality around setting tint color for ProgressBar in a background compatible way, as a developer I want a ProgressBar subclass that takes care of that automatically.

---

(Coming out from the review of https://github.com/schedjoules/android-event-discovery-sdk/pull/185)

This code
```
if (Build.VERSION.SDK_INT < 21)
        {
            // on older Version the progressbar won't pick up the accent color, so we have to set it manually
            mViews.schedjoulesEventListProgressBar.getIndeterminateDrawable()
                    .setColorFilter(getResources().getColor(R.color.schedjoules_colorAccent),
                            PorterDuff.Mode.MULTIPLY);
        }
``` 
can be improved to this:
```
DrawableCompat.setTint(DrawableCompat.wrap(views.schedjoulesEventDetailsHorizontalActionsProgressbar.getIndeterminateDrawable()),
                        ContextCompat.getColor(getContext(), R.color.schedjoules_colorAccent)
```
(API level check could be kept as well, to not do the tinting unnecessarily)
And applied in the ProgressBar subclass, which we can then reference in xml.

Hardcoding the color to accentColor is probably good enough.

dmfs:
Make sure to use this progress bar in all places (also see #211)

lemonboston:
See also https://github.com/schedjoules/android-event-discovery-sdk/pull/208#issuecomment-295753043

PR #208, opened 2017-04-14
Reference colors instead of hard-coding them. Implements #206

dmfs:
As a by-product this also adds a dark theme.
Also reformat all resource files.

Please make sure attribute "Wrap always" is enabled in your XML code style.


lemonboston:
Wouldn't it be possible and enough to just have the partners override a style resource to change the theme? Like:
```
<style name="SchedJoules_EventDiscovery_Theme" parent="ShedJoules_Theme.Dark"/>
```
In their styles.xml. (We would have the same declaration in our own with the default Light)

`SchedJoules_Theme.Dark.NoActionBar` - Does `NoActionBar` in the name have a purpose? I mean, is it possible to customize the presence of the action bar? I guess not and in this case, can we remove from the name, so simply end with `.Dark`?

> Please make sure attribute "Wrap always" is enabled in your XML code style.

I use `dmfs` code style and had 3 out of the 4 dropdowns on Android tab set to 'wrap_always', now set the 4th one as well, but I still have whitespace diff after formatting the changed files. Can we maybe put the code style xml under version control?

lemonboston:
I leave this review until the possible style override question is resolved.

dmfs:
Some partners want to be able to switch themes at runtime, hence we need to be able to support dynamic themes.

dmfs:
@lemonboston please check again. Sorry, that I had to force push because I made those changes halfway through the rebase on master, which means there is no separate commit.

dmfs:
Partners can now change just a few values of the theme  by overriding `SchedJoules_BaseTheme.Light.Custom` (and/or `SchedJoules_BaseTheme.Dark.Custom`) in their styles like so:

```xml
    <style name="SchedJoules_BaseTheme.Light.Custom">
         <item name="colorAccent">#ff00ff</item>
    </style>
```
They don't have to override all the values nor should they have to deal with SDK level specific settings.

In general we should encourage partners not to override any values, but changing the accent color is probably an acceptable use case for this. This is preferred over overriding the actual `schedjoules_colorAccent` color resource.

lemonboston:
Details screen icon colors are not changed with the `colorAccent` override, because that’s defined like this:
```
<color name="schedjoules_content_icon">@color/schedjoules_colorAccent</color>
```
So it seems we can’t cross reference our colors like this which are used as theme colors as well..
Maybe we should not have `schedjoules_colorPrimary` and others defined, just ‘hardcode’ the color in the theme definition, to avoid referencing them in any other way than `?attr:colorPrimary`.


```
<style name="SchedJoules_Theme" parent="SchedJoules_Theme.Light"/>
```
This looks as a circular hierarchy definition, not sure why Android allows it.. To avoid confusion, could we have a different name for them? For example
```
<style name="SchedJoules_AppTheme" parent="SchedJoules_Theme.Light"/>
```

Why do we need to keep both `SchedJoules_BaseTheme.Light.NoActionBar` and `SchedJoules_BaseTheme.Light`? Won’t all our Activities be without ActionBar because of using Toolbar?

`styles.xml`
Shouldn’t these be defined in the `theme_light/dark.xml`s?
```
    <style name="SchedJoules_AppTheme.AppBarOverlay.Dark" parent="ThemeOverlay.AppCompat.Dark.ActionBar"/>

    <style name="SchedJoules_AppTheme.PopupOverlay" parent="ThemeOverlay.AppCompat.Light"/>

```
`SchedJoules_AppTheme.PopupOverlay` seems to be used for location picker toolbar, it would have a menu popup with incorrect color probably.
I am actually amazed that you could put this alternate theme configuration together so fast.
But I worry a bit we miss something, so I wonder whether we should create a task to double-check all style/them usage in all layouts.

`EventListFragment`
```
TypedValue typedValue = new TypedValue();
        getActivity().getTheme().resolveAttribute(R.attr.schedjoules_dropdownArrow, typedValue, true);
```
Nice that we can use theme dependent resources with those attributes.
Here, instead of the two arrow drawable which only differs in `fillColor` and making the drawable an attr, could we use a color attribute and one drawable?
Like: http://stackoverflow.com/a/12706305
Then we don’t even need code change in `EventListFragment`, I suppose.

I see this is an extra attribute in the dark theme:
```
<item name="android:textColorPrimaryInverse">@color/schedjoules_white</item>
```
Just curiosity, where is it used? Do we not have to set it for the Light as well?

This whole theme separation looks really good to me overall.

About the runtime Theme configuration, as discussed in detail on Slack, I think a global parameter would be better instead of making sure to add theme to all Intents. So I would suggest implementing that in a separate story and removing current solution from this pull request.

dmfs:
First of all, please use the review function to comment directly in the code. That makes it much easier to respond to and to follow the individual issue threads. Also you (and I) don't have to repeat the code lines you refer to.

I've fixed the details icons, they use the accent color now, there is no need for a dedicated color here. If there was, we would have to create an attribute.

Hard coding colors is an option but only for colors that we use just once. In any case our new policy is to not use color resources unless there is a **very** good reason to it (like in case of the expanded details header title, which is always white). We should watch out for color resources during reviews.

There is no circular dependency in 
```xml
<style name="SchedJoules_Theme" parent="SchedJoules_Theme.Light"/>
```
`SchedJoules_Theme.Light` does not inherit from `SchedJoules_Theme` because it has a different parent theme. Anyway I've renamed `SchedJoules_Theme` to ``SchedJoules_Theme.Default` to make it more clear.

`SchedJoules_BaseTheme.Light.NoActionBar` exists because it sets SDK Level specific values. Without it we would have to maintain a copy of `SchedJoules_BaseTheme.Light` in `values` and another one in `values-v21`, although we just want to add two more values. So the `NoActionBar` theme adds some SDK level specific stuff.

The two dark styles you refer two are actually not theme specific and used in both themes. You're right, they shouldn't be used in the location picker, so I removed these entries. They are now solely used for the expanded details toolbar.

We certainly should test this branch on all Android versions, but our theme is not that complex that we could have missed much. Even if there are a few things that can not be customized properly (because we still sue hard coded values) I'm not worried, because these will pop up over time and we can fix them. For now it's sufficient if the light and dark theme look as expected.

Regarding the one drawable with an attribute color: That's not possible. Drawables can only use color resources (at least on SDK level <21). They don't support "colors by reference". That's also why we can't create drawables with "accent color" in XML. It's said that this is supported since SDK Level 21, but we still need to support older SDK levels, so there is no benefit if we use this technique.

Not sure about the inverse color. At some point I needed it to fix an issue, but I can't remember what it was and it doesn't seem to be necessary anymore. I've added a comment. I need to check on Android 4 again to be sure it's not in use anymore.

I guess we agree, that there is no way to persist the selected theme permanently (because all identifiers we have might be subject to change). So a partner would have to configure it at runtime. This also means, if a partner uses switchable themes, he needs to ensure that our activities can't be called directly from the outside, otherwise it might use the wrong theme. Iit doesn't matter how he would actually set the theme, he always has to intercept any call and run his own code first (to make sure the theme is set).

For now I want to keep the current method until we come across some real issues with it. The current solution works is simple and efficient and I don't see any benefit in making it more complex. I've made the startActivity methods final, because they certainly should not be overridden though.

eventdiscovery-sdk/src/main/res/values/theme.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <!-- This sets the default theme to light -->
+    <style name="SchedJoules_Theme.Default" parent="SchedJoules_Theme.Light"/>
```
lemonboston, line 5: >There is no circular dependency in..
SchedJoules_Theme.Light does not inherit from SchedJoules_Theme because it has a different parent theme. Anyway I've renamed SchedJoules_Theme to ``SchedJoules_Theme.Default` to make it more clear.

Since hierarchy can be defined with both the parent attribute and by using dot in the name, I think this is still confusing.
https://developer.android.com/guide/topics/ui/themes.html#Inheritance
I've found the article where I read about it, it has a good description and reasoning:
http://blog.danlew.net/2014/11/19/styles-on-android/  (Implicit vs. Explicit Parenting)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/Color.java
```diff
@@ -0,0 +1,87 @@
...
+
+
+/**
+ * Represents the Color referenced by a specific attribute. The only supported method is {@link #intValue()}.
+ * <p>
+ * TODO: with Android O you can create meaningful instances of {@link android.graphics.Color}. We probably can use this as the superclass then.
+ *
+ * @author Marten Gajda
+ */
+public final class Color extends Number
```
lemonboston: I like this idea to model color, but it's not really a number for me. Could we maybe have a dedicated interface?
With a single method currently, something like `int rgb()` - not sure what representation is that int is actually, but a method name that refers to that representation.
And then we could also implement `ContextCompat.getColor(context, colorRes)` underneath. So maybe 2 implementations, `ResourceColor` and `ThemeAttributeColor`.

eventdiscovery-sdk/src/main/res/values/colors_light.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Colors of the light theme -->
+
+    <color name="schedjoules_colorPrimary">@color/schedjoules_white</color>
```
lemonboston, line 5: Shouldn't we then get rid of all these 'intermediaries' that are used as theme attributes? Because if theme attributes are overriden by partner, then these 'detached' references are still used using the old color. Just as with the details icons. So in order to avoid that, shouldn't we remove these and instead reference the theme attribute wherever these are used?
(This what I wanted to say in the previous post, where I confusingly said 'hardcoding' the theme colors.)
This could of course be a separate task.

eventdiscovery-sdk/src/main/res/values-v21/theme_light.xml
```diff
@@ -1,8 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <style name="SchedJoules_AppTheme.NoActionBar">
+
+    <style name="SchedJoules_BaseTheme.Light.NoActionBar">
```
lemonboston, line 4: My motivation for the notes about these `NoActionBar` themes is to possibly take it out from the name as it seems to be the default now and probably forever, to not have the legacy ActionBar, so just have simpler base theme names.
But I understand that extra level is needed to add v21 version. Not sure what would be better naming than, though..
I see also that only 
```
<item name="android:windowDrawsSystemBarBackgrounds">true</item>
<item name="android:statusBarColor">@android:color/transparent</item>
```
are extra for v21. 
So just as an unsure input for what could be in values/styles.xml:
```
<style name="SchedJoules_Theme.Light.Actual"/>
       
    <style name="SchedJoules_Theme.Light" parent="Theme.AppCompat.Light">
        <item name="windowActionBar">false</item>
        <item name="windowNoTitle">true</item>
        <item name="colorPrimary">@color/schedjoules_colorPrimary</item>
        <item name="colorPrimaryDark">@color/schedjoules_colorPrimaryDark</item>
        <item name="colorAccent">@color/schedjoules_colorAccent</item>
```
And then v21 has that 2 extra attributes for `SchedJoules_Theme.Light.Actual`. The name is not very good, I know, so current verision is okay, too, for me.

eventdiscovery-sdk/src/main/res/drawable/schedjoules_ic_arrow_drop_down_light.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="@dimen/schedjoules_toolbar_title_font_size"
+        android:height="@dimen/schedjoules_toolbar_title_font_size"
+        android:viewportWidth="24.0"
+        android:viewportHeight="24.0">
+    <path
+            android:fillColor="@color/schedjoules_dark_text_primary"
```
lemonboston, line 7: > Regarding the one drawable with an attribute color: That's not possible. Drawables can only use color resources (at least on SDK level <21). They don't support "colors by reference". That's also why we can't create drawables with "accent color" in XML. It's said that this is supported since SDK Level 21, but we still need to support older SDK levels, so there is no benefit if we use this technique.

I just tried this with `android:fillColor="?attr/colorAccent"` and it worked for pre-Lollipop emulators as well.

lemonboston (commented):
> please use the review function to comment directly in the code. 

Ok, I will do this from now on.

I haven't 'tested' the app yet, to check through the UI on different Android version, I will comment again when I am done with that.

eventdiscovery-sdk/src/main/res/values-v21/theme_light.xml
lemonboston: Nevermind this, I missed to see that there is `SchedJoules_Theme.Light` which is the parent of the `NoActionBar` version, so there is a short name for the top theme.

lemonboston:
Just for aesthetic reasons, could we change the demo app accent color to something else? :)
Like this blue from material colors: #2196F3 I think that looks nice.

lemonboston:
On Android 4.4 with the white theme, so overridden accent color to purple, the loading circle on the list and the loading 'ribbon' on the details is still green.
Linking this note to https://github.com/schedjoules/android-event-discovery-sdk/issues/211 if you'd prefer fixing it there.

lemonboston:
Haven't found anything else during 'testing', finished the review.

Should we maybe have a different default accent color for the dark theme?

eventdiscovery-sdk/src/main/res/drawable/schedjoules_ic_arrow_drop_down_light.xml
dmfs: If I set
```
    <path
            android:fillColor="?android:attr/textColorSecondary"
            android:pathData="M6,6l6,6 6,-6z"/>
```
I get the lint warning:
> Using theme references in XML drawables requires API level 21 (current min is 14)

Maybe the Android SDK can do something smart about app attributes, but it definitely doesn't work with system attributes.

eventdiscovery-sdk/src/main/res/values/theme.xml
dmfs: I don't agree with his reasoning. If I inherit from `ThemeA.B` I don't care if `ThemeA` is the actual parent of `ThemeA.B` or not. I want to inherit from `ThemeA.B` and not from `ThemeA`.

FWIW, the support library uses the dot notation with explicit parenting for pretty much every style. 

Also, for our partners it doesn't matter at all, because they should only use `SchedJoules_Theme.Default`, `SchedJoules_Theme.Light` or `SchedJoules_Theme.Dark` and just as outlined above, they don't care if these themes inherit from `SchedJoules_Theme` or not.
Everything else is internal business and I think our themes are still simple enough to not get it wrong.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/Color.java
dmfs: Good point. Will do.

dmfs:
I don't really care about the accent color of the demo that much. It's really just for demo purposes. I'll change it to blue.

The progress bar tint issue on Android 4 should be fixed by #192.

For now we keep the accent color of the dark theme. I'll ask Jurjen for suggestions.

dmfs:
@lemonboston please review again.

eventdiscovery-sdk/src/main/res/values/colors_light.xml
dmfs: Yes, we'll take care of that later.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/Color.java
```diff
@@ -0,0 +1,33 @@
...
+ * @author Marten Gajda
+ */
+public interface Color
+{
+    /**
+     * Returns the color as an int in argb format.
+     *
+     * @return The color value.
+     */
+    int argb();
```
lemonboston: `@ColorInt` to here as well?

lemonboston (commented):
Regarding theme parenting: Ok

Created task for removing named color references that are used as theme attributes: https://github.com/schedjoules/android-event-discovery-sdk/issues/218

> If I set
> ```
>     <path
>             android:fillColor="?android:attr/textColorSecondary"
>             android:pathData="M6,6l6,6 6,-6z"/>
> ```
> I get the lint warning:
> > Using theme references in XML drawables requires API level 21 (current min is 14)
>
> Maybe the Android SDK can do something smart about app attributes, but it definitely doesn't work with > system attributes.

I see, so do you have a concern using app attribute? Why not use it if it works?
I found these about the topic:
https://android.jlelse.eu/android-developers-we-ve-been-using-themes-all-wrong-eed7755da586
_Unfortunately, in all previous to Lollipop, there was a bug that prevented you from using ?attr/ in your drawables_
https://issuetracker.google.com/issues/36941443

There are merge conflicts now, if you can resolve, squash and rebase, I could do a final round of check.

(Technical note: I couldn't reply to previous review comments now - there is simply no box there for it - so that's why I copied them here.
Edit2: Now there are boxes, so probably it was because I was in the process of a next review.)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/Color.java
dmfs: Good catch!

dmfs:
Done.

Regarding the app attributes in drawables, since it's officially broken on Android 4.x we're not going to use it.

lemonboston:
I couldn't find any official source on this attribute support in drawables. Did you maybe?
There is that bug where they say it was released in L-preview. But it does work in pre-Lollipop for me, and I wonder that this might be something that is handled simply by the newer build tools. Anyway, it would be nice to use it if possible, but I understand that if it has any risk, then we should not.

Otherwise, I am done with the review, looks good to me.

PR #214, opened 2017-04-18
PR for 213: Set up project code style settings

lemonboston:
I've applied the `dmfs.xml` code style setup as project setup and it changes `codeStyleSettings.xml` so from now on if we choose 'Project' settings in Studio code style settings, it should be the same for everybody.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -3,6 +3,7 @@
   <component name="ProjectCodeStyleSettingsManager">
     <option name="PER_PROJECT_SETTINGS">
       <value>
+        <option name="FIELD_NAME_PREFIX" value="m" />
         <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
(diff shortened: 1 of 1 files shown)


.idea/codeStyleSettings.xml
```diff
@@ -63,169 +76,41 @@
           </extensions>
         </Objective-C-extensions>
         <XML>
+          <option name="XML_ATTRIBUTE_WRAP" value="2" />
           <option name="XML_KEEP_LINE_BREAKS" value="false" />
+          <option name="XML_KEEP_LINE_BREAKS_IN_TEXT" value="false" />
```
dmfs: please change to `true`.

dmfs (changes requested):

.idea/codeStyleSettings.xml
lemonboston: Updated, squashed, rebased.

dmfs (approved):

Issue #204, opened 2017-04-12
Google Play services upgrade message


dmfs:
This is how it looks like when you open the details view on a device with an outdated Google Play services app:

![device-2017-04-12-234658](https://cloud.githubusercontent.com/assets/1614987/24981135/8b17ccca-1fdb-11e7-8147-af9e35bbb213.png)

The problem: you can't press the button to update the app, because it launches Google Maps

dmfs:
This should become obsolete when #207 has been implemented.

lemonboston:
Just putting the note here that it was suggestion to use MapView inside our own Fragment: https://github.com/schedjoules/android-event-discovery-sdk/pull/215#pullrequestreview-33903042

PR #215, opened 2017-04-19
PR for 207: Fix map fragment crash by adding it from code instead of xml

lemonboston:
@dmfs I looked up [MapView docs](https://developers.google.com/android/reference/com/google/android/gms/maps/MapView) and saw this instruction:
_"Users of this class must forward all the life cycle methods from the Activity or Fragment.."_
So it looks cumbersome to use `MapView` directly.

On the other hand, in the SO post you posted, they say the crash is general for fragments defined in xml, but doesn't happen when added from code: http://stackoverflow.com/a/31755655

So I fixed it by just adding the `SupportMapFragment` from code to a `FrameLayout` holder.


dmfs (approved):
It fixes the crash, but I'd still like to see that we use a `MapView` instead. I suggest to create our custom `MapFragment` as part of #204.

Issue #194, opened 2017-04-11
Clickable area of actions too small


dmfs:
The clickable area of the actions in the details view should cover the entire item with and height.

The directions action has the correct width, but is not high enough.

SHARE and ADD TO CALENDAR should span the entire width of the action item.

All actions are supposed to have a height of 64 dp (directions may be higher if necessary).

PR #227, opened 2017-04-25
PR for 194: Adjust event details actions clickable areas

lemonboston:
Implements #194.

Known issue/question: For multiline direction action, the icon is vertically center aligned to the height, not aligned with the first line (as venue name).
I left it out for 2 reasons:
1. I wanted to ask it again if we want this behavior for sure and not aligning between 1st and 2nd line, like here in the Google Calendar:
device-2017-04-25-183319
Ask opinions on ui-ux chat?
2. To keep this changeset small, I would do the refactor to re-use this behavior between venue name and directions in a separate task, possibly creating a `TextView` subclass. (For either version: aligning to 1st line or between 1st and 2nd line)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/ActionView.java
```diff
@@ -51,6 +52,13 @@ public ActionView(SchedjoulesViewEventDetailsActionBinding binding)
     }
 
 
+    public ActionView(SchedjoulesViewEventDetailsActionDirectionsBinding directionsBinding)
+    {
+        mRoot = directionsBinding.getRoot();
```
dmfs: better delegate to a private constructor like so
```
   this(directionsBinding.getRoot(), directionsBinding.schedjoulesEventDetailsDirectionsLabel);
```
Same goes for the other constructor of course.

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/ActionView.java
lemonboston: Right, it's better that way. I've updated it.

lemonboston:
Squashed, rebased, ready for merge.

Created #229 for the icon alignment.

dmfs (approved):

Issue #175, opened 2017-04-03
Location picker query update animation: same items no longer stay in place


lemonboston:
For example typing 's' after 'Am' to be 'Ams', the item 'Amsterdam' should not disappear and reappear, but stay in place. This worked before, a recent change broke it.

---

The cause is the `Clickable` decorator for `ListItem`, they no longer can be compared with `equals()` with that. 
Idea to fix this is to introduce item identification and item content identification for `ListItem`s, like:
```
interface ListItem
{
   // existing methods
  String id();
  String contentId();
}
```
So `DiffUtil` can use `id()` for `areItemsTheSame()` and `contentId()` for `areContentsTheSame()` and apply correct animations accordingly. This may also allow to update items with progress indication or change click actions, etc.

lemonboston:
As suggested and discussed using `Equalable` instead of `String` for the return types above is better, the plan is to do it that way.

PR #230, opened 2017-04-26
PR for 175: Fix location picker list update animation for same items

lemonboston:
Implements #175.

One small thing: I've added this `@Frequent` annotation to mark methods that are called often, to pay attention to optimization. I would like it to be added to implementations automatically, like `@NonNull` is added when I generate overriden methods in the IDE - it even warns on the interface if not all implementation added that. Not sure yet how to set up `@Frequent` similarly.

As mentioned on chat, I left out `contentId()` because I tried it and it didn't work as expected:
For example when I updated an item with 'text1' to 'text2', then first the 'text1' flashed and kept being 'text1', then flashed again and become 'text2'. So I didn't want to spend more time on it to investigate how this really works, since we don't need this at the moment, also I was not sure what's the best design for this actually (we could combine `id()` and `contentId()` to a type to make for easier `ListItem` implementations).

`id()`s are not implemented in the event list item types now, because we use `FlexibleAdapter` there and don't use `DiffUtil` so they wouldn't be used anyway.


dmfs (approved):

dmfs:
Note, I'm not convinced about this `@Frequent` annotation, mostly because it makes assumptions about how a method is going to be used and not about the method itself.
I mean pretty much every method could potentially called often. Also, this depends heavily on algorithms this is used in.

Issue #235, opened 2017-04-28
Android serialization design


lemonboston:
In order to have less boilerplate and code duplication around Android serialization (and thus help maintainability, allow to work more efficiently with this later) as a developer, I would like a better architectural design to handle this aspect.
---

PR #236, opened 2017-04-28
Android serialization mini-framework. #235

lemonboston:
Initial version submitted for discussion and improvements.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListActivity.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListActivity.java
@@ -21,6 +21,7 @@ import android.os.Bundle;
 
 import com.schedjoules.eventdiscovery.R;
 import com.schedjoules.eventdiscovery.framework.common.BaseActivity;
+import com.schedjoules.eventdiscovery.framework.serialization.commons.NestedBundle;
 
 
 /**
@@ -41,7 +42,7 @@ public final class EventListActivity extends BaseActivity
         if (savedInstanceState == null)
         {
             getSupportFragmentManager().beginTransaction().add(R.id.schedjoules_activity_content,
-                    EventListFragment.newInstance(getIntent().getExtras())).commit();
+                    EventListFragment.newInstance(new NestedBundle(getIntent()).get())).commit();
         }
     }
 
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListFragment.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListFragment.java
@@ -56,6 +56,8 @@ import com.schedjoules.eventdiscovery.framework.eventlist.view.EventListLoadingI
 import com.schedjoules.eventdiscovery.framework.eventlist.view.EventListMenu;
 import com.schedjoules.eventdiscovery.framework.locationpicker.LocationPickerPlaceSelection;
 import com.schedjoules.eventdiscovery.framework.locationpicker.SharedPrefLastSelectedPlace;
+import com.schedjoules.eventdiscovery.framework.serialization.Keys;
+import com.schedjoules.eventdiscovery.framework.serialization.commons.OptionalArgument;
 import com.schedjoules.eventdiscovery.framework.utils.FutureServiceConnection;
 import com.schedjoules.eventdiscovery.framework.utils.InsightsTask;
 import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
@@ -69,9 +71,6 @@ import org.dmfs.rfc5545.DateTime;
 
 import eu.davidea.flexibleadapter.FlexibleAdapter;
 import eu.davidea.flexibleadapter.items.IFlexible;
-
-import static com.schedjoules.eventdiscovery.framework.EventIntents.EXTRA_GEOLOCATION;
-import static com.schedjoules.eventdiscovery.framework.EventIntents.EXTRA_START_AFTER_TIMESTAMP;
 
 
 /**
@@ -300,28 +299,17 @@ public final class EventListFragment extends BaseFragment implements EventListMe
 
     private GeoLocation location()
     {
-        if (getArguments() != null && getArguments().containsKey(EXTRA_GEOLOCATION))
+        if (new OptionalArgument<>(Keys.GEO_LOCATION, getArguments()).isPresent())
         {
             // TODO Save location and update Toolbar title with name when Event Discovery for input geo-location is actually supported
-            return getArguments().getParcelable(EXTRA_GEOLOCATION);
+            throw new UnsupportedOperationException("Discovery for a geo location not supported yet.");
         }
-        else
-        {
-            return new SharedPrefLastSelectedPlace(getContext()).get().geoLocation();
-        }
+        return new SharedPrefLastSelectedPlace(getContext()).get().geoLocation();
     }
 
 
     private DateTime startAfter()
     {
-        if (getArguments() != null && getArguments().containsKey(EXTRA_START_AFTER_TIMESTAMP))
-        {
-            long startAfterTimeStamp = getArguments().getLong(EXTRA_START_AFTER_TIMESTAMP, 0);
-            return new DateTime(startAfterTimeStamp);
-        }
-        else
-        {
-            return DateTime.nowAndHere();
-        }
+        return new OptionalArgument<>(Keys.DATE_TIME_START_AFTER, this).value(DateTime.nowAndHere());
     }
 }
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/DateTimeBox.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/DateTimeBox.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.serialization.boxes;
+
+import android.os.Parcel;
+
+import com.schedjoules.eventdiscovery.framework.serialization.core.Box;
+
+import org.dmfs.rfc5545.DateTime;
+
+import java.util.TimeZone;
+
+
+/**
+ * {@link Box} for {@link DateTime}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class DateTimeBox implements Box<DateTime>
+{
+
+    private final DateTime mDateTime;
+
+
+    public DateTimeBox(DateTime dateTime)
+    {
+        mDateTime = dateTime;
+    }
+
+
+    @Override
+    public DateTime content()
+    {
+        return mDateTime;
+    }
+
+
+    @Override
+    public int describeContents()
+    {
+        return 0;
+    }
+
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags)
+    {
+        TimeZone timeZone = mDateTime.getTimeZone();
+        dest.writeString(timeZone == null ? null : timeZone.getID());
+        dest.writeLong(mDateTime.getTimestamp());
+        dest.writeInt(mDateTime.isAllDay() ? 1 : 0);
+
+    }
+
+
+    public static final Creator<DateTimeBox> CREATOR = new Creator<DateTimeBox>()
+    {
+        @Override
+        public DateTimeBox createFromParcel(Parcel in)
+        {
+            String timeZoneId = in.readString();
+            long timestamp = in.readLong();
+            boolean isAllDay = in.readInt() != 0;
+
+            TimeZone timeZone = timeZoneId == null ? null : TimeZone.getTimeZone(timeZoneId);
+            DateTime dateTime = new DateTime(timeZone, timestamp);
+            if (isAllDay)
+            {
+                dateTime = dateTime.toAllDay();
+            }
+            return new DateTimeBox(dateTime);
+        }
+
+
+        @Override
+        public DateTimeBox[] newArray(int size)
+        {
+            return new DateTimeBox[size];
+        }
+    };
+}
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Argument.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Argument.java
@@ -20,10 +20,13 @@ package com.schedjoules.eventdiscovery.framework.serialization.commons;
 import android.app.Activity;
 import android.content.Intent;
 import android.os.Bundle;
+import android.support.v4.app.Fragment;
 
 import com.schedjoules.eventdiscovery.framework.serialization.core.Key;
 import com.schedjoules.eventdiscovery.framework.utils.factory.AbstractLazy;
 import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
+
+import org.dmfs.optional.NullSafe;
 
 
 /**
@@ -55,5 +58,11 @@ public final class Argument<T> extends AbstractLazy<T>
     public Argument(Key<T> key, Activity activity)
     {
         this(key, activity.getIntent());
+    }
+
+
+    public Argument(Key<T> key, Fragment fragment)
+    {
+        this(key, new NullSafe<>(fragment.getArguments()).value(Bundle.EMPTY));
     }
 }
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Argument.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Argument.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.serialization.commons;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Bundle;
+
+import com.schedjoules.eventdiscovery.framework.serialization.core.Key;
+import com.schedjoules.eventdiscovery.framework.utils.factory.AbstractLazy;
+import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
+
+
+/**
+ * Represents an argument in a parameterizable container (Bundle or Intent practically).
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class Argument<T> extends AbstractLazy<T>
+{
+    public Argument(final Key<T> key, final Bundle bundle)
+    {
+        super(new Factory<T>()
+        {
+            @Override
+            public T create()
+            {
+                return new OptionalArgument<>(key, bundle).value();
+            }
+        });
+    }
+
+
+    public Argument(Key<T> key, Intent intent)
+    {
+        this(key, new NestedBundle(intent).get());
+    }
+
+
+    public Argument(Key<T> key, Activity activity)
+    {
+        this(key, activity.getIntent());
+    }
+}
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.serialization.commons;
+
+import android.os.Bundle;
+
+import com.schedjoules.eventdiscovery.framework.serialization.core.Box;
+import com.schedjoules.eventdiscovery.framework.serialization.core.Boxable;
+import com.schedjoules.eventdiscovery.framework.serialization.core.FluentBuilder;
+import com.schedjoules.eventdiscovery.framework.serialization.core.Key;
+
+import org.dmfs.optional.Optional;
+
+
+/**
+ * {@link FluentBuilder} for {@link Bundle}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class BundleBuilder implements FluentBuilder<Bundle>
+{
+    private final Bundle mBundle;
+
+
+    public BundleBuilder()
+    {
+        this(Bundle.EMPTY, false);
+    }
+
+
+    public BundleBuilder(Bundle bundle)
+    {
+        this(bundle, true);
+    }
+
+
+    private BundleBuilder(Bundle bundle, boolean clone)
+    {
+        mBundle = clone ? new Bundle(bundle) : bundle;
+    }
+
+
+    @Override
+    public Bundle build()
+    {
+        return new Bundle(mBundle);
+    }
+
+
+    @Override
+    public <T> FluentBuilder<Bundle> with(Key<T> key, Box<T> box)
+    {
+        Bundle bundle = new Bundle(mBundle);
+        bundle.putParcelable(key.name(), box);
+        return new BundleBuilder(bundle, false);
+    }
+
+
+    @Override
+    public <T> FluentBuilder<Bundle> with(Key<T> key, Optional<Box<T>> optBox)
+    {
+        if (!optBox.isPresent())
+        {
+            return this;
+        }
+        return with(key, optBox.value());
+    }
+
+
+    @Override
+    public <T extends Boxable> FluentBuilder<Bundle> withOptionalBoxable(Key<T> key, Optional<T> optBoxable)
+    {
+        if (!optBoxable.isPresent())
+        {
+            return this;
+        }
+        return with(key, optBoxable.value().boxed());
+    }
+
+}
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
@@ -83,7 +83,7 @@ public final class BundleBuilder implements FluentBuilder<Bundle>
 
 
     @Override
-    public <T extends Boxable> FluentBuilder<Bundle> withOptionalBoxable(Key<T> key, Optional<T> optBoxable)
+    public <T extends Boxable> FluentBuilder<Bundle> withBoxable(Key<T> key, Optional<T> optBoxable)
     {
         if (!optBoxable.isPresent())
         {
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
@@ -99,7 +99,7 @@ public final class IntentBuilder implements FluentBuilder<Intent>
 
 
     @Override
-    public <T extends Boxable> FluentBuilder<Intent> withOptionalBoxable(Key<T> key, Optional<T> optBoxable)
+    public <T extends Boxable> FluentBuilder<Intent> withBoxable(Key<T> key, Optional<T> optBoxable)
     {
         if (!optBoxable.isPresent())
         {
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
@@ -21,7 +21,6 @@ import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
 
-import com.schedjoules.eventdiscovery.framework.serialization.Keys;
 import com.schedjoules.eventdiscovery.framework.serialization.core.Box;
 import com.schedjoules.eventdiscovery.framework.serialization.core.Boxable;
 import com.schedjoules.eventdiscovery.framework.serialization.core.FluentBuilder;
@@ -75,7 +74,7 @@ public final class IntentBuilder implements FluentBuilder<Intent>
     @Override
     public Intent build()
     {
-        mIntent.putExtra(Keys.NESTED_BUNDLE, mNestedBundleBuilder.build());
+        mIntent.putExtra(NestedBundle.NESTED_BUNDLE_KEY, mNestedBundleBuilder.build());
         return new Intent(mIntent);
     }
 
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.serialization.commons;
+
+import android.content.Intent;
+import android.net.Uri;
+import android.os.Bundle;
+
+import com.schedjoules.eventdiscovery.framework.serialization.Keys;
+import com.schedjoules.eventdiscovery.framework.serialization.core.Box;
+import com.schedjoules.eventdiscovery.framework.serialization.core.Boxable;
+import com.schedjoules.eventdiscovery.framework.serialization.core.FluentBuilder;
+import com.schedjoules.eventdiscovery.framework.serialization.core.Key;
+
+import org.dmfs.optional.Optional;
+
+
+/**
+ * {@link FluentBuilder} for {@link Intent}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class IntentBuilder implements FluentBuilder<Intent>
+{
+    private final Intent mIntent;
+    private final FluentBuilder<Bundle> mNestedBundleBuilder;
+
+
+    public IntentBuilder(Intent intent)
+    {
+        this(intent, new BundleBuilder(new NestedBundle(intent).get()));
+    }
+
+
+    public IntentBuilder()
+    {
+        this(new Intent());
+    }
+
+
+    public IntentBuilder(String action)
+    {
+        this(new Intent(action));
+    }
+
+
+    public IntentBuilder(String action, Uri uri)
+    {
+        this(new Intent(action, uri));
+    }
+
+
+    private IntentBuilder(Intent intent, FluentBuilder<Bundle> nestedBundleBuilder)
+    {
+        mIntent = new Intent(intent);
+        mNestedBundleBuilder = nestedBundleBuilder;
+    }
+
+
+    @Override
+    public Intent build()
+    {
+        mIntent.putExtra(Keys.NESTED_BUNDLE, mNestedBundleBuilder.build());
+        return new Intent(mIntent);
+    }
+
+
+    @Override
+    public <T> FluentBuilder<Intent> with(Key<T> key, Box<T> box)
+    {
+        return new IntentBuilder(mIntent, mNestedBundleBuilder.with(key, box));
+    }
+
+
+    @Override
+    public <T> FluentBuilder<Intent> with(Key<T> key, Optional<Box<T>> optBox)
+    {
+        if (!optBox.isPresent())
+        {
+            return this;
+        }
+        return with(key, optBox.value());
+    }
+
+
+    @Override
+    public <T extends Boxable> FluentBuilder<Intent> withOptionalBoxable(Key<T> key, Optional<T> optBoxable)
+    {
+        if (!optBoxable.isPresent())
+        {
+            return this;
+        }
+        return with(key, optBoxable.value().boxed());
+    }
+
+}
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
@@ -20,18 +20,26 @@ package com.schedjoules.eventdiscovery.framework.serialization.commons;
 import android.content.Intent;
 import android.os.Bundle;
 
-import com.schedjoules.eventdiscovery.framework.serialization.Keys;
 import com.schedjoules.eventdiscovery.framework.utils.factory.AbstractLazy;
 import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
 
 
 /**
- * Represents the nested {@link Bundle} in an {@link Intent}.
+ * Lazy access to an {@link Intent}s nested {@link Bundle} if it's present, or to a new empty {@link Bundle} if it's not.
+ * <p>
+ * (Note: Nested {@link Bundle} are used inside {@link Intent}s as a practice to avoid crashes when the system would try to
+ * unparcel custom classes outside of the app.)
  *
  * @author Gabor Keszthelyi
  */
 public final class NestedBundle extends AbstractLazy<Bundle>
 {
+    /**
+     * The key for the nested Bundle Intent extra.
+     */
+    public static final String NESTED_BUNDLE_KEY = "intent.extra.NESTED_BUNDLE";
+
+
     public NestedBundle(final Intent intent)
     {
         super(new Factory<Bundle>()
@@ -39,7 +47,7 @@ public final class NestedBundle extends AbstractLazy<Bundle>
             @Override
             public Bundle create()
             {
-                Bundle nestedBundle = intent.getBundleExtra(Keys.NESTED_BUNDLE);
+                Bundle nestedBundle = intent.getBundleExtra(NESTED_BUNDLE_KEY);
                 return nestedBundle == null ? new Bundle() : nestedBundle;
             }
         });
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.serialization.commons;
+
+import android.content.Intent;
+import android.os.Bundle;
+
+import com.schedjoules.eventdiscovery.framework.serialization.Keys;
+import com.schedjoules.eventdiscovery.framework.utils.factory.AbstractLazy;
+import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
+
+
+/**
+ * Represents the nested {@link Bundle} in an {@link Intent}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class NestedBundle extends AbstractLazy<Bundle>
+{
+    public NestedBundle(final Intent intent)
+    {
+        super(new Factory<Bundle>()
+        {
+            @Override
+            public Bundle create()
+            {
+                Bundle nestedBundle = intent.getBundleExtra(Keys.NESTED_BUNDLE);
+                return nestedBundle == null ? new Bundle() : nestedBundle;
+            }
+        });
+    }
+
+}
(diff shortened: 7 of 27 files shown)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/Box.java
```diff
@@ -0,0 +1,31 @@
...
+import android.os.Parcelable;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public interface Box<T> extends Parcelable
+{
+
+    T content();
```
dmfs: As mentioned in Slack, please consider to add a `Context` parameter so we can potentially box stuff that requires a `Context` to be restored.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/SimpleKey.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.schedjoules.eventdiscovery.framework.serialization.commons;
+
+import com.schedjoules.eventdiscovery.framework.serialization.core.Key;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class SimpleKey<V> implements Key<V>
```
dmfs: How about the name `StringKey`? The key is derived from a `String` so that would make sense to me.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
```diff
@@ -0,0 +1,47 @@
...
+public final class NestedBundle extends AbstractCachingOptional<Bundle>
+{
+    public NestedBundle(final Intent intent)
+    {
+        super(new Factory<Optional<Bundle>>()
+        {
+            @Override
+            public Optional<Bundle> create()
+            {
+                return new NullSafe<>(intent.getBundleExtra(Keys.NESTED_BUNDLE));
```
dmfs: How about wrapping this with `Obligatory` so the `NestedBundle` is always present (falling back to an empty bundle)?
```
return new Obligatory(new NullSafe<>(intent.getBundleExtra(Keys.NESTED_BUNDLE)), Bundle.EMPTY);
```
(optionally use `new Bundle()` instead of `Bundle.EMPTY` if the result should be mutable).

That way you wouldn't have to write
```
NestedBundle(getIntent()).value(null)
```
above. It could just be
```
NestedBundle(getIntent()).value()
```
If the `NestedBundle`, by contract, is always present.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Extra.java
```diff
@@ -0,0 +1,40 @@
...
+{
+
+    public Extra(final Key<T> key, final Intent intent)
+    {
+        super(new Factory<Optional<T>>()
+        {
+            @Override
+            public Optional<T> create()
+            {
+                if (intent != null)
```
dmfs: I guess this could be replaced by
```
return new Argument<T>(key, new NestedBundle(intent).value());
```

or even better, if `Argument` would take an `Optional<Bundle>` 
```
return new Argument<T>(key, new NestedBundle(intent)); 
```

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Argument.java
```diff
@@ -0,0 +1,59 @@
...
+import org.dmfs.optional.Optional;
+import org.dmfs.optional.Present;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public class Argument<T> extends AbstractCachingOptional<T>
+{
+    // TODO Worth implementing here, not using the abstract class
```
dmfs: Totally agree!

Also consider to accept `Optional<Bundle>` as a parameter ideally using lazy evaluation. So it could look like this:
```
   private final Key<T> mKey;
   private final Optional<Bundle> mBundle;
 
  public Argument(final Key<T> key, final Bundle bundle) 
  {
     this(key, new Present<>(bundle));
  }

  public Argument(final Key<T> key, final Optional<Bundle> bundle) 
  {
      mKey = key;
      mBundle = bundle;
  }

  boolean isPresent()
  {
     return mBundle.value(Bundle.EMPTY).contains(key.name());
  }

  // etc...
```

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Extra.java
dmfs: Btw. don't check `intent` for `null`. If you pass a `null` Intent you're most likely doing it wrong and the code should not swallow this error silently.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Argument.java
dmfs: Also note, it should be safe to assume that the passed `Bundle` is never `null`. Otherwise the caller should wrap it with `NullSafe<>(bundle)`. Let's see if we can eliminate `null` checks altogether.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/DateTimeBox.java
```diff
@@ -0,0 +1,80 @@
...
+    public int describeContents()
+    {
+        return 0;
+    }
+
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags)
+    {
+        dest.writeLong(mDateTime.getTimestamp());
```
dmfs: you should also write time zone and maybe the all-day flag otherwise the reconstructed value might be incorrect.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListFragment.java
```diff
@@ -300,28 +299,17 @@ public void onDestroy()
...
-        if (getArguments() != null && getArguments().containsKey(EXTRA_START_AFTER_TIMESTAMP))
-        {
-            long startAfterTimeStamp = getArguments().getLong(EXTRA_START_AFTER_TIMESTAMP, 0);
-            return new DateTime(startAfterTimeStamp);
-        }
-        else
-        {
-            return DateTime.nowAndHere();
-        }
+        return new Argument<>(Keys.DATE_TIME_START_AFTER, getArguments()).value(DateTime.nowAndHere());
```
dmfs: It would be nice if we can find a simple mechanism to evaluate the fallback value lazily. There are cases in which the fallback value consumes many resources and it's a shame if we create it when it's never used. I just had this case somewhere else.
Something like the `Obligatory` decorator that takes a fallback factory instead of a specific fallback value could solve this.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListActivity.java
```diff
@@ -41,7 +42,7 @@ protected void onCreate(Bundle savedInstanceState)
         if (savedInstanceState == null)
         {
             getSupportFragmentManager().beginTransaction().add(R.id.schedjoules_activity_content,
-                    EventListFragment.newInstance(getIntent().getExtras())).commit();
+                    EventListFragment.newInstance(new NestedBundle(getIntent()).value(null))).commit();
```
dmfs: see my comment on `NestedBundle` below. The `null` fallback looks completely wrong to me.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BasicFluentIntent.java
```diff
@@ -0,0 +1,68 @@
...
+
+
+    public BasicFluentIntent()
+    {
+        this(new LinkedList<Parameter<Intent>>());
+    }
+
+
+    @Override
+    public void into(Intent intent)
```
dmfs: despite the fact that this `Parameter` interface looks procedural, this particular case could be solved more elegantly, without mutating an existing `Intent`.

I'm thinking of something like an `IntentBuilder`. Like so:
```
Intent newIntent = new IntentBuilder(oldIntent).withExtra(key, box).intent();
assertNotSame(oldIntent, newIntent);
```
or even like this (imitating the other constructors of an Intent):
```
Intent intent = new IntentBuilder(Intent.ACTION_VIEW, url).withExtra(key, box).intent();
```

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/Parameter.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.schedjoules.eventdiscovery.framework.serialization.core;
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public interface Parameter<T>
+{
+
+    void into(T container);
```
dmfs: I don't like this concept of mutating an existing container. Maybe consider using a builder style pattern like 

```
public interface Parametrized<T>
{
    T updated();

    Parametrized<T> withParameter(Key<V> key, V value);
}
```

Which would be used like this
```
Bundle newBundle = new Parametrized(oldBundle).withParameter(key, value).updated();
```

The builder should be immutable of course.

Not sure if that would work for all uses cases that you have in mind.

Also, the name `Parametrized` is not very good, because it would indicate a decorator, which is not the case. Maybe `BuildAdapter` would be more accurate (but it contains the pattern name, which is not very nice). I would call instances just like `BundleBuilder` or `IntentBuilder` omitting the "Adapter" part.

dmfs (changes requested):
done with the first round

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/Box.java
lemonboston: (Note: I start to comment here, but will probably make updates for the other comments)

One thing that have come up now is that `Box<T>` can be used inside `Parcelable` implementations as well, and we don't have `Context` there in the `CREATOR`. 
Example:
```
    public void writeToParcel(Parcel dest, int flags)
    {
        dest.writeParcelable(mDelegate, flags);
        dest.writeParcelable(new DateTimeBox(mDate), flags);
    }

        public Dated createFromParcel(Parcel in)
        {
            ClassLoader classLoader = getClass().getClassLoader();
            return new Dated((
                    (DateTimeBox) in.readParcelable(classLoader)).content(),
                    (EventFilter) in.readParcelable(classLoader));
        }
```
(This is in current `serialization.temp.parcelable.Dated` - will write about those later..).

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/SimpleKey.java
lemonboston: Ok, changed it to `StringKey`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
lemonboston: I thought about this `Obligatory` in the weekend and now I think that it can be misleading to use. I mean that wherever we need an `Obligatory<T>`, don't we actually need a `T`?
That's what its name means I think, so it's no longer an `Optional<T>`. If we pass it around as `Optional<T>` subsequent code will see it that way, providing its own defaults or routing on `isPresent()`, while it really should not at that point, because that is no longer a valid program flow.

As I see the motivation in this case, it would be to have a lazy `Bundle` that always evaluates to something. I think that's not the same as `Optional<Bundle>`.
Because `Bundle` is not an interface we can't do it in an implementation, so I guess the alternative to
```
Bundle bundle = new NestedBundle(getIntent()).value(null)
or
Bundle bundle = new NestedBundle(getIntent()).value(new Bundle());
```
would be a factory then..  but I am not sure that's worth it.

The contract that nested bundle would always present be - would than mean we also make the contract that all Fragments has a non-null argument?
The null check in `Argument` is because of that, so that `fragment.getArguments()` can be null, so just to be convenient to use `new Arugment(.. fragment.getArguments()). 
We could set up these contracts to eliminate nulls as much as possible, then we need to keep to that, making sure to add empty bundles in the parameterless Fragment factory methods.

Let me know if we should do this.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Extra.java
lemonboston: Right, simplified it and also removed the `null` check for `intent` - you're right that can never really be `null` normally.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/Box.java
dmfs: Hmm ... good point. I guess one solution would be to use the DateTimeBox internally instead of the date. Like so:

```
  private Box<DateTime> mDateBox;

  public Dated(DateTime date, EventFilter delegate)
  {
     this(new DateTimeBox(date), delegate);
  }

  private Dated(DateTimeBox datebox, EventFilter delegate)
  {
     mDateBox = datebox;
     mDelegate = delegate;
  }

  public void writeToParcel(Parcel dest, int flags)
    {
        dest.writeParcelable(mDelegate, flags);
        dest.writeParcelable(mDateBox, flags);
    }

        public Dated createFromParcel(Parcel in)
        {
            ClassLoader classLoader = getClass().getClassLoader();
            return new Dated(
                    (DateTimeBox) in.readParcelable(classLoader),
                    (EventFilter) in.readParcelable(classLoader));
        }
```

Also, how would that look if `Dated` doesn't implement `Box` but `Boxable` instead? Shouldn't that fix it too?

dmfs: ah, scratch that. It would still not have a `Context`. So that's not an option. Ok I guess we can't do this then.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Argument.java
lemonboston: It comes back to the same question, I think, whether we agree on a contract that all Fragments has to have an argument.
Just because of convenient usage:
```
// Convenient:
Event event = new Argument(Keys.EVENT, getArguments()).value();
// Non-convenient:
Event event = new Argument(Keys.EVENT, new NullSafe<>(getArguments())).value();
```
I think we should avoid having the need to write down the second one, so either we keep null check in `Argument` (either explicit or using `NullSafe`) or we introduce a contract to always have argument for Fragment.
Right?

lemonboston: About implementing the `Optional`s directly instead of using the abstract class - I will rewrite all these classes later as it makes sense to optimize them since they will be core, often used classes.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/DateTimeBox.java
lemonboston: Ok, I've added them.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListFragment.java
lemonboston: As mentioned above, I think we are talking about a factory here indeed, to create an object lazily based on conditions. Optional would be a misuse for this, I think.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListActivity.java
lemonboston: Discussed above.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Argument.java
lemonboston: Oh, I just realized that a paremeterless Fragment will not have code to extract the argument! :)
Silly me. So we can safely assume that bundle is non-null when we actually want to use it.
I will update code accordingly.

lemonboston: Done. Just ignore my previous comments on the `null` topic.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/Parameter.java
lemonboston: I really like your idea to not mutate the container but return a modified copy! That was a basic problem here..
And I see there are copy constructors for both `Intent` and `Bundle`.

I am not sure though how well the above, combined `Parametrized<T>` would work..  
Wouldn't it work well only if we implement it once? For example, please check current `BundleParameter` and `ParcelableBundleParameter`. The 2nd one has input params `Key` and `Parcelable`, not `Key` and `Box`.
So I am wondering if it's better to keep them separate, the common implementations for the actual parameter entry and the builder.
So what do you think about this?:
```
interface Param<C>
{
    C updated(C originalCarrier);
}

interface FluentParam<C> extends Param<C>
{
    <T> FluentParam withParam(Key<T> key, Box<T> box);
}
```
Names can change of course. Maybe no need for the general `Fluent*`, we can have specific for `Intent`s with `withExtra()` and for `Bundles` with `withArg()`.

Usage could be this:
```
Bundle bundle = new BundleBuilder()
    .withParam(key1, box1)
    .withParam(key2, box2)
    .updated(new Bundle());

or for single param:
Bundle bundle = new BundleParam(key, box).updated(new Bundle());
Bundle bundle = new ParcelableBundleParam(key, parc).updated(new Bundle());
```

lemonboston: Although, your usage looks better..  maybe it's better that way then, with one implementation and writing a bit more when we need to put the Box together.

lemonboston: I will do it the way you suggested with dedicated interfaces for Intent and Bundle builder.

lemonboston:
I've added examples of how to do the serialization for the `EventFilter` example,
so in general when
- we have our own type definition, and
- we will surely need to access internals for serialization, and
- we want decoration to still work

Four options for it:
1. Don't define serialization on the interface, just on the implementor classes -> isn't good because decorator couldn't serialize this way
2. The interface extends `Parcelable` -> this will be the good one, I think. We don't need `Box` at this point, doesn't add anything. We put it into a `Box` with `ParcelableBox` when we build the `Intent`/`Bundle`.
3. The interface implements `Box<T>` -> as mentioned above, no benefit over 2., but we have an extra method `content()` which always `return this`.
4. The interface implements `Boxable<T>` which has method `Box<T> boxed()` -> this looked like the best first, because it has some resemblance with what we did for `Equalable ListView.id()`, to make sure decoration works. But it seems this case is different. Both 3. and 4. just adds extra methods to the class (`content()`, `content()+boxed()`).

Small note: I learned that we can't implement `Parcelable` in an inner class, cause it can't have the creator static field. 

They are in the `serialization.temp` package, please check if you agree that we just need to go with
```
interface OurType extends Parcelable
```
in these cases.

dmfs:
You can make inner classes `Parcelable` if they are `static`. So you need to pass any state or optionally the object to parcel to the constructor. It's slightly more verbose but it's essentially what happens under the hood with a non static inner class.

This is the idea:
```java
public final class Dated implements EventFilter
{
   private final EventFilter mDelegate;
   private final DateTime mDateTime;

   / / constructors here

    @Override
    public Box<EventFilter> boxed()
    {
      return new DatedBox(this);
      // or probably better:
      return new DatedBox(mDelegate, mDateTime);
    }

  // other methods here …

  private final static class DatedBox implements Box<Dated>
  {
     private final Dated mDated;

     public DatedBox(Dated dated)
     {
        mDated = dated;
     }

   /*
     optional constructor to pass individual values instead of the Dated object.
     The advantage is that Java doesn't have to create "access(…)" methods

     With this method DatedBox doesn't even have to be a private inner class and
     could just be a regular public toplevel class.
  */
     public DatedBox(EventFilter filterDelegate, DateTime datetime)
     {
        mFilterDelegate = filterDelegate;
        mDateTime = datetime;
     }

     // all the parcelable stuff here access mDated methods like mDated.mDateTime
  }

}
```

dmfs:
Also note that in your case 4, the `EventFilter` interface wouldn't extend `Box` but just `Boxable`. So the only method added is `boxed()`. Whereas `Parcelable` adds 2 methods and a static field.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/Parameter.java
dmfs: Not sure what you mean by "dedicated interfaces for Intent and Bundle builder.", but you probably can use the same interface for both of them. The only thing that would be different is the name of the actual build method I guess.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/temp/parcelable/Everything.java
```diff
@@ -0,0 +1,67 @@
...
+
+import android.os.Parcel;
+
+import com.schedjoules.eventdiscovery.discovery.EventDiscovery;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+class Everything implements EventFilter
```
dmfs: It just occurred to me that `AllEventsFilter` or `NoFilter` (depending on how you look at it) might be better names.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
dmfs: If you say it like that you could call every method a "conditional factory", but that's not the point.

Try to see it like this. An `Obligatory` like this:
```java
new Obligatory(optional, default);
```
is the declarative equivalent of these two expressions:
```java
optional.isPresent()? optional : new Present<>(default);

new Present(optional.value(default));
```
Also, note that `Obligatory` is a decorator to an existing `Optional`, it's not a type nor a "standalone class". I agree that there wouldn't be any point in an `Obligatory<T>` interface, because we could just use `<T>` instead. 

This `Optional` design is different from Java 8's `Optional` which is rather procedural in that you often use it to describe an algorithm ("if … is not present then fall back to …") whereas our `Optional` (and this `Obligatory` in particular) can be used to describe the *result* in a declarative way ("the value of …, falling back to … if necessary").

My idea is that we should not have to use `isPresent()` very often when we use optionals. For instance, if I pass an `Optional<Bundle>` to an `Extra`, the `Extra` should be absent if either the `Bundle` is absent or if it doesn't contain the extra. I don't want to call `isPresent` on the optional Bundle before I create the extra.

```java
  Optional<X> y = new Extra(new NestedBundle(intent));
```

Also it would be great if the same would work with a "builder":

```java
  Intent newIntent = new IntentBuilder("action").withExtra(key, new Extra(new NestedBundle(oldIntent))).build().
```
In that case `newIntent` should only contain the extra if it was present in `oldIntent`.

Or to give you a more practical example. `BaseActivity` contains this code:
```java
Extra<Integer> themeExtra = new Extra<>(Keys.THEME, getIntent());
if (themeExtra.isPresent())
{
      new IntentParameter<>(Keys.THEME, new IntBox(themeExtra.value())).into(intent);
}
return intent;
```

I think it should look like this:
```java
return new IntentBuilder(intent).withExtra(Keys.THEME, new Extra<>(Keys.THEME, getIntent())).build();
```
Resulting in an intent that contains the theme extra if and only if it was present in the original intent (i.e. build an Intent with whatever the current Intent contains as the theme)

So maybe also consider to update the builder adapter interface below with a method like this
```java
  // add an Optional extra only if present
  <T> BuilderAdapter withExtra(Key<T> key, Optional<T> value);
```
(names are subject to change).


So, to get back to the original issue: You don't have to apply `Obligatory` in `NestedBundle` but if you don't you should be more clear when you use it.
Your original code was 
```java
NestedBundle(getIntent()).value(null);
```
Which doesn't make any sense. Why would it default to `null`? Is it always expected to be present then just call `value()` (yes, we can't create a custom Bundle, that's a pitty, but even if we could, the interface would be unsuitable for this) if the `NestedBundle` is allowed to be absent, either use a proper fallback (if the method or constructor you were passing this to requires a value) or allow the constructor to take an `Optional` if absent values are valid parameters too.

lemonboston:
I've added those variations on the `Boxable<T>` implementation. Although we need to implement the same things in all 3 main options, the fact that we can move it to another top level class with `Boxable<T>`, thus separating concerns does look better indeed. So 'real' logic is not hidden among  serialization boilerplate.
So I am on board now to use that one (`Dated_v3` + `DatedBox_v3`). Thanks.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/temp/parcelable/Everything.java
lemonboston: Updated, although this is just temporary code for that here, I will remove them.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
lemonboston: > new NestedBundle(getIntent()).value(null);

This was only replicating the previous usage (`getIntent().getExtras()`) where it could be null as well, but I agree it's better to always provide a Bundle arg for a Fragment which can have arguments.

>```
><T> BuilderAdapter withExtra(Key<T> key, Optional<T> value);
>```

We have `Key<T>` and `Box<T>` as parameters, we can't use general `<T>` or `Optional<T>` cause we don't know how to box it.

So that's why this won't work either:
> ```
> return new IntentBuilder(intent).withExtra(Keys.THEME, new Extra<>(Keys.THEME, getIntent())).build();
> ```

But I've add a `withCopiedExtra(Key<T> key, Intent intent)` to the builder, that works.

Btw, `NestedBundle` should really not be used normally, that's one aim of this design is to hide that boilerplate. It's internal to how we put extras into Intent. The usage in `EventListActivity` was only needed to take out the intent extras and forward as fragment bundle.

So I've added the initial version for the builders, please check them out.

I will write to you on Slack about the Obligatory and Optional usage topic.

dmfs: Hmm. good point.
We could do that at least for `Boxable`:
```java
<T extends Boxable> BuilderAdapter withExtra(Key<T> key, Optional<T> value);
```

And for Optional Boxes
```java
<T> BuilderAdapter withExtra(Key<T> key, Optional<Box<T>> value);
```

That doesn't help here though. But I'm wondering if there isn't another way. Like a `BoxedExtra` that returns the boxed Object instead of the unboxed one. `BoxedExtra<T>` would be an `Optional<Box<T>>` and could be used with the method above. Or am I missing something?

I'm not a big fan of the `withCopiedParam` method. The use case looks way to limited to me. It doesn't work that well when I want to copy a Fragment Extra into an Intent or vice versa.

lemonboston: Both of the above methods couldn't be added like this because they have the same erasure,
but I've added the second one and that `BoxedExtra<T>`. So it can work like that.
I've removed the `withCopiedParam`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/DateTimeBox.java
```diff
@@ -0,0 +1,91 @@
...
+    public int describeContents()
+    {
+        return 0;
+    }
+
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags)
+    {
+        dest.writeString(mDateTime.getTimeZone().getID());
```
dmfs: Note that the current `DateTime` implementation is quite old and will return `null` as the time zone for all-day events.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+    public <T> FluentBuilder<Bundle> withParam(Key<T> key, Box<T> box)
+    {
+        return withParam(key, new Present<>(box));
+    }
+
+
+    @Override
+    public <T> FluentBuilder<Bundle> withParam(Key<T> key, Optional<Box<T>> optBox)
+    {
+        Bundle bundle = new Bundle(mBundle);
```
dmfs: if `optBox` is not present you can just return `this`, because the bundle is not going to be modified in that case.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Extra.java
```diff
@@ -0,0 +1,32 @@
...
+import com.schedjoules.eventdiscovery.framework.utils.factory.Factory;
+import com.schedjoules.eventdiscovery.framework.utils.optionals.AbstractCachingOptional;
+
+import org.dmfs.optional.Optional;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class Extra<T> extends AbstractCachingOptional<T>
```
dmfs: As you already mentioned, this should be called `OptionalExtra`

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Extra.java
```diff
@@ -0,0 +1,32 @@
...
+{
+    // TODO Implement Optional here, not use abstract class
+
+
+    public Extra(final Key<T> key, final Intent intent)
```
dmfs: Maybe consider creating a constructor that takes an activity or even context instead of an intent.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+    public <T> FluentBuilder<Intent> withParam(Key<T> key, Box<T> box)
+    {
+        return withParam(key, new Present<>(box));
+    }
+
+
+    @Override
+    public <T> FluentBuilder<Intent> withParam(Key<T> key, Optional<Box<T>> optBox)
+    {
+        FluentBuilder<Bundle> bundleBuilder = optBox.isPresent() ? mNestedBundleBuilder.withParam(key, optBox.value()) : mNestedBundleBuilder;
```
dmfs: Same here, if `optBox` is not present just return `this`

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/FluentBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author Gabor Keszthelyi
+ */
+public interface FluentBuilder<R>
+{
+    R build();
+
+    <T> FluentBuilder<R> withParam(Key<T> key, Box<T> box);
+
+    <T> FluentBuilder<R> withParam(Key<T> key, Optional<Box<T>> optBox);
+}
```
dmfs: I still think that having a method like
```
  <T extends Boxable> FluentBuilder<R> withParam(Key<T> key, Optional<T> box); 
```
could make things easier. You could give it another name to avoid signature conflicts.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/FluentBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Interface for builders that use {@link Key} and {@link Box} to add parameters to their built result object.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface FluentBuilder<R>
+{
+    R build();
+
+    <T> FluentBuilder<R> withParam(Key<T> key, Box<T> box);
```
dmfs: maybe consider to rename `withParam` to just `with`. the signature of the method makes pretty clear what it means.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/DateTimeBox.java
lemonboston: Thanks, I've fixed it by simply passing null as `TimeZone` to the `DateTime` constructor in this case.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
lemonboston: I was not sure about this, what would be the reasonable contract to this kind of immutable object 'with' method with optional parameter, should it return new instance for absent value or not.
But in practice, I see that we will probably not use the same builder to create multiple Bundles or Intents, so I changed it. I will also add note to the javadoc.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
lemonboston: Updated this one as well.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Extra.java
lemonboston: Done.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
dmfs: That's the beauty and the whole point of immutable objects. If we don't change anything we don't have to return a new instance and in fact there is no risk in returning the same instance and no benefit in returning a new instance. For the caller it doesn't matter at all.

Even if we use the same builder for multiple Bundles or Intents, this won't cause any problems, because the object is immutable. Any call to `with` that would change the state will create a new instance.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Extra.java
lemonboston: Added the constructor with `Activity`.
(`Context` doesn't have generic way to retrieve its `Intent`, right?)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/FluentBuilder.java
lemonboston: Good idea, thanks, I like these concise names here. (Updated)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/FluentBuilder.java
lemonboston: Added with `withOptionalBoxable` name (not the best possibly, let me know if you have better idea)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/Extra.java
dmfs: Right, a `Context` may not have an `Intent` (just think of the application context). I think only `Activity` and `Services` have `Intents`. So forget about the context idea ;-)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/FluentBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface FluentBuilder<R>
+{
+    R build();
+
+    <T> FluentBuilder<R> with(Key<T> key, Box<T> box);
+
+    // TODO tell in javadoc that absent will not return new instance
```
dmfs: That won't be necessary, because it really doesn't matter since all instances must be immutable anyway.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
```diff
@@ -0,0 +1,107 @@
...
+            return this;
+        }
+        return new IntentBuilder(mIntent, mNestedBundleBuilder.with(key, optBox.value()));
+    }
+
+
+    @Override
+    public <T extends Boxable> FluentBuilder<Intent> withOptionalBoxable(Key<T> key, Optional<T> box)
+    {
+        return with(key, box.isPresent() ? new Present<Box<T>>(box.value().boxed()) : Absent.<Box<T>>absent());
```
dmfs: That looks a bit more complicated that it has to IMO.
Why do you call `with(Key<T> key, Optional<Box<T>> optBox)`?
how about calling `with(Key<T> key, Box<T> box)` like so?
```java
    if (box.isPresent())
    {
        return this;
     }
     return with(key, box.value().boxed());
```

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+        bundle.putParcelable(key.name(), optBox.value());
+        return new BundleBuilder(bundle);
+    }
+
+
+    @Override
+    public <T extends Boxable> FluentBuilder<Bundle> withOptionalBoxable(Key<T> key, Optional<T> box)
+    {
+        return with(key, box.isPresent() ? new Present<Box<T>>(box.value().boxed()) : Absent.<Box<T>>absent());
```
dmfs: see my comment in `IntentBuilder`

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BoxedExtra.java
```diff
@@ -0,0 +1,50 @@
...
+public final class BoxedExtra<T> extends AbstractCachingOptional<Box<T>>
+{
+    public BoxedExtra(final Key<T> key, final Intent intent)
+    {
+        super(new Factory<Optional<Box<T>>>()
+        {
+            @Override
+            public Optional<Box<T>> create()
+            {
+                Bundle nestedBundle = new NestedBundle(intent).value(Bundle.EMPTY);
```
dmfs: I would remove these redundant variables and write a one-liner. I think having these variables doesn't improve readability.

lemonboston: (This code is not present like this anymore)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
lemonboston: You're right! :)

However `BundleBuilder` was not really immutable, I knew I would need to review them, I've done  it now and refactored a bit. So the mutable `Bundle` is copied now in both inbound and outbound.
`IntentBuilder` is changed at one point because of this.

(I will do one more full review of the branch shortly, check these ones, too.)

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/IntentBuilder.java
lemonboston: True, I've simplified them.

lemonboston:
I've realized that we don't need separate classes for the parameter extraction from `Bundle` and `Intent`, just extra constructors. Added these general ones as `Parameter` and `OptionalParameter`.
By the way, I've also added the mandatory version as `Lazy`.

So I would suggest we keep only the general ones. The question is the name. And although I didn't like this `Argument` name for the `Bundle` version, if we keep only one, I would vote for that.
I've looked up once before what's the difference of 'parameter' and 'argument' and they say here for example:
http://stackoverflow.com/a/156778/4247460
that 'parameter' is more like the declaration, formal one and 'argument' is the actual value.
So for the extracting side I think 'argument' fits well, we have an actual value there to represent, not a general parameter.
So I suggest renaming current `Parameter` and `OptionalParameter` to `Argument` and `OptionalArgument` and remove the others.
What do you think?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+
+    public BundleBuilder()
+    {
+        mBundle = new Bundle();
+    }
+
+
+    private BundleBuilder(Bundle bundle, String nonUsedParam)
```
dmfs: I'd prefer a slightly different pattern to have this internal non-cloning constructor.
Consider this:

```
   public BundleBuilder() 
   {
       // we don't need to clone Bundle.EMPTY, because we don't try to mutate it
       this(Bundle.EMPTY, false);
   }

   public BundleBuilder(Bundle bundle) 
   {
       this(bundle, true);
   }

   private BundleBuilder(Bundle bundle, boolean clone)
   {
      mBundle = clone ? new Bundle(bundle) : bundle;
   } 
```

dmfs:
I've fine with "Argument"

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/BundleBuilder.java
lemonboston: You're right, that's nicer, thanks. I've updated it.

lemonboston:
I've squashed and rebased and finalized everything, ready for review.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/factory/Lazy.java
```diff
@@ -26,6 +26,10 @@
 {
     /**
      * Returns the instance which is created on first access.
+     * <p>
+     * Note:
+     * If <code>T</code> is mutable, the returned object is always the same instance.
+     * If <code>T</code> is immutable, the instance can be different (as there is no need for it to be the same).
```
dmfs: These two notes are pointless. The important point is that the "identity" of the result is the same on every call. It doesn't matter if the result is mutable or not. I think the definition of `Lazy` already makes clear how it behaves. Besides, this method doesn't get any parameters, so the result is expected to be the same on every call (since the `Lazy` instance is supposed to be immutable).

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/factory/AbstractLazy.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Gabor Keszthelyi
+ */
+public abstract class AbstractLazy<T> implements Lazy<T>
+{
+    private final Lazy<T> mDelegate;
+
+
+    public AbstractLazy(Factory<T> factory)
+    {
+        mDelegate = new SimpleLazy<>(factory);
```
dmfs: Why does `AbstractLazy` build on `SimpleLazy`? Wouldn't it make more sense to be the other way round?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/FluentBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+     * Same as {@link #with(Key, Box)} but with taking an {@link Optional<Box>}.
+     * If that is absent, no change is applied.
+     */
+    <T> FluentBuilder<R> with(Key<T> key, Optional<Box<T>> optBox);
+
+    /**
+     * Same as {@link #with(Key, Box)} but with taking an {@link Optional<Boxable>} to take the {@link Box} from.
+     * If that is absent, no change is applied.
+     */
+    <T extends Boxable> FluentBuilder<R> withOptionalBoxable(Key<T> key, Optional<T> optBoxable);
```
dmfs: I think `withBoxable` should be sufficient for a name. `withOptionalBoxable` is certainly "more correct", but it's quite long and cumbersome. For consistency the other 2 methods could be called `withBox` then. I think the `Optional` doesn't have to be part of the name in this case, since it feels more like sort of an attribute rather than a type.
Does that make sense?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/OptionalArgument.java
```diff
@@ -0,0 +1,95 @@
...
+    public OptionalArgument(Key<T> key, Bundle bundle)
+    {
+        mKey = key;
+        mBundle = bundle;
+    }
+
+
+    public OptionalArgument(Key<T> key, Fragment fragment)
+    {
+        this(key, fragment.getArguments());
```
dmfs: This will cause an NPE if the fragment has no arguments `Bundle`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
```diff
@@ -0,0 +1,48 @@
...
+{
+    public NestedBundle(final Intent intent)
+    {
+        super(new Factory<Bundle>()
+        {
+            @Override
+            public Bundle create()
+            {
+                Bundle nestedBundle = intent.getBundleExtra(Keys.NESTED_BUNDLE);
+                return nestedBundle == null ? new Bundle() : nestedBundle;
```
dmfs: This class is a bit weird. It says "Represents the nested {@link Bundle} in an {@link Intent}.", first of all, it doesn't "represent" a `Bundle`, and secondly, what does it really represent if there is no nested Bundle in the intent?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/factory/Lazy.java
lemonboston: Ok, I removed them.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/factory/AbstractLazy.java
lemonboston: So true. I've changed it.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/core/FluentBuilder.java
lemonboston: Yes, I agree that `withBoxable` is sufficient, I've changed it.
I am not sure about `withBox` though, I think the method adds a parameter which is the key and the value together, so key is important as well. `withBox` would only refer to the value. I know it would be to be consistent with `withBoxable`, but I suspect `with` will be used much more often, so I'd like it to be that concise.
So as I see `withParameter`, `withParam` , `with` are appropriate names, `withBox` not so much.
So I would rather treat `withBoxable` as a necessary technical exception on this interface and keep the other two as `with`.
But let me know if you'd prefer them to be changed.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/OptionalArgument.java
lemonboston: Right, good catch, thanks. Updated.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/commons/NestedBundle.java
lemonboston: I've changed the description, please check if it makes more sense now.
I've also moved the nested bundle key to this class as it needs to be referenced by the general `IntentBuilder` as well, so it can't really be a project specific value.

lemonboston:
Ready for check again.

dmfs (approved):
Ok, let's see it in action.

PR #237, opened 2017-04-28
PR for 209: Fix AppBar casting shadow on StatusBar

lemonboston:
> The reason is probably the transparent background color of the status bar. Activities (or "root-fragments") that don't use a collapsingtoolbarlayout should probably set the status bar background to a solid primaryColorDark

Yes, that seemed to be the cause, AppBar probably casts shadow on the status bar. I've found the issue asked here:
http://stackoverflow.com/questions/34891756/toolbars-shadow-on-status-bar-for-lollipop
I've tried to handle it in theme xmls but hit a dead end after a while: if partners would use multiple themes dynamically, not just the overriden `Default`, it wouldn't work.
(Otherwise working version about that approach is on branch `bugs/209-appbar-shadow-on-statusbar`)

So I looked up whether we can change status bar color from code and we can:
http://stackoverflow.com/a/26749343/4247460 
So I used this and it turned out to be a much simpler version than trying to do it with themes.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/common/StatusBar.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/common/StatusBar.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


dmfs:
Please resolve the conflicts

lemonboston:
Conflicts resolved.

dmfs (approved):

PR #241, opened 2017-05-02
PR for 192: Backwards compatible tinted ProgressBar

lemonboston:
Implements #192.

I've change the accent color of the demo app from blue to orange because apparently the `ProgressBar` horizontal default color on pre-Lollipop is a very similar blue. (It take away about half an hour..). So to avoid missing problems with the accent color, I've changed it.

The new tinted horizontal ProgressBar animation on pre-Lollipop looks different for some reason compared to how it looks like without it. So originally the line is moving fast all across the width, now it starts slowly from the left. Please check it out if it's okay like this or should I spend time to try to fix this. It doesn't look too bad imo and it may not be worth spending time on older version look fine tuning.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/AccentColoredProgressBar.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/AccentColoredProgressBar.java
@@ -51,8 +51,7 @@ public final class AccentColoredProgressBar extends ProgressBar
         Drawable indeterminateDrawable = getIndeterminateDrawable();
         if (indeterminateDrawable != null)
         {
-            setIndeterminateDrawable(TintedDrawable.tint(indeterminateDrawable, new AttributeColor(getContext(), R.attr.colorAccent)));
-
+            setIndeterminateDrawable(TintedDrawable.tinted(indeterminateDrawable, new AttributeColor(getContext(), R.attr.colorAccent)));
         }
     }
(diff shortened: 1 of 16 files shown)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/widgets/AccentColoredProgressBar.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+
+    private void tintIndeterminateDrawable()
+    {
+        Drawable indeterminateDrawable = getIndeterminateDrawable();
+        if (indeterminateDrawable != null)
+        {
+            setIndeterminateDrawable(TintedDrawable.tint(indeterminateDrawable, new AttributeColor(getContext(), R.attr.colorAccent)));
```
dmfs: didn't you suggest to call `mutate` on the drawable to avoid side effects?

lemonboston: `TintedDrawable.tint(..)` calls `mutate()`.
It also calls `DrawableCompat.wrap()` first which according to the documentation is need to tint the `Drawable` but I remember it working without that. So I am not certain about that, but probably best to keep that, too.

I've renamed the method `tint()` to `tinted()` to better express that the returned object is a changed, different one.

dmfs (approved):

Issue #210, opened 2017-04-18
Words are linkified incorrectly in event description


lemonboston:
![device-2017-04-18-170424](https://cloud.githubusercontent.com/assets/1077972/25137806/175e18de-2459-11e7-8d6f-e4b643ecca8d.png)

http://stackoverflow.com/questions/30528825/android-autolink-is-too-aggressive
http://stackoverflow.com/questions/43384870/text-with-dot-displayed-as-hyperlink

dmfs:
On some devices this also applies to numbers. On my phone every number with at least 2 digits is linkified as a phone number, which is pretty annoying.

dmfs:
What we need is custom link patterns: http://www.indelible.org/ink/android-linkify/
@wilfredstegeman which patterns should we linkify? Some of the options are

 * full URLs (i.e. "http://...")
 * host names (i.e. URLs with missing http or https scheme, like "example.com" or also "example.com/abc"). This is error prone and produces false positives as you can see above. One way to deal with that is to do this in two steps, first linkify without this and show the result, but also start a linkify background task that checks if the linked phrases refer to existing host names (i.e. do a DNS lookup) and replace the result when this has finished.
 * email addresses
 * phone numbers, we would need to apply certain constraints when we decide what's a phone number (like a minimal length of 5-8 digits or so).
 * twitter mentions (i.e. "@xxxx")

lemonboston:
As a first iteration and improvement on this we could leave out the host names (web address without "http(s)://"), and also phone numbers and twitter from the list above. So basically linkify only full URLs and email addresses. To keep it simpler and make a step forward.
 
@dmfs @wilfredstegeman What do you think?

Then we could create separate tasks for each of the other cases if needed.

dmfs:
@lemonboston yes that's exactly what we should do, start with the easy ones.

PR #247, opened 2017-05-03
Change event description plain text linkification to only linkify full web urls and email addresses. #210

lemonboston:
@dmfs This is not submitted for final review, more like for discussing questions.
It contains initial solution for web urls starting with "http" and for email addresses.

One question is: do we need to linkify htmls as well or do the links in them come inside appropriate tags already and not in plain text content?
(Note: I've briefly tried to apply current linkification to the html formatted `Spanned` - it removed other formatting like the header style. But maybe it can be solved, not sure at this point.)

**A bigger question**: if server side will need to process htmls anyway, maybe at an increased sophistication in the future, could this be handled there as well? I mean to **make all event description html**, for example just by adding a `<p>..</p>` around plain texts. And take care of all the linkification with adding the appropriate tags.

If you look at current code in `StrictHttpUrlLinkified`, and for example this link:
https://mathiasbynens.be/demo/url-regex
than I think this question becomes more apparent. It's not a trivial thing to do, so if server could do this platform independently with html tags, I wonder whether that would be better. 

Other TODOs
* The current two `Spannable` decorators for email and url only work in the current order, figure out why is that
* There is a `LinkifyCompat` class, check if it was better to use that

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/linkify/CompleteWebUrlLinkifying.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/linkify/CompleteWebUrlLinkifying.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.utils.linkify;
+
+import android.support.v4.text.util.LinkifyCompat;
+import android.text.Spannable;
+import android.text.util.Linkify;
+import android.util.Patterns;
+
+import com.schedjoules.eventdiscovery.framework.utils.spanned.AbstractSpannable;
+
+import java.util.regex.Matcher;
+
+
+/**
+ * A {@link Spannable} that linkifies web urls that starts with "http" or "Http" in the input {@link Spannable}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class CompleteWebUrlLinkifying extends AbstractSpannable
+{
+
+    private static final Linkify.MatchFilter MATCH_FILTER = new HttpMatchFilter();
+    private static final Linkify.TransformFilter TRANSFORM_FILTER = new ToLowerCaseHttp();
+
+
+    public CompleteWebUrlLinkifying(final Spannable input)
+    {
+        super(new com.schedjoules.eventdiscovery.framework.utils.factory.Factory<Spannable>()
+        {
+            @Override
+            public Spannable create()
+            {
+                LinkifyCompat.addLinks(input, Patterns.WEB_URL, null, MATCH_FILTER, TRANSFORM_FILTER);
+                return input;
+            }
+        });
+    }
+
+
+    /**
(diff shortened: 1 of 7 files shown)


lemonboston:
Examples for the html links that could be added by the server:
```
<a href="http://google.com">google.com</a>
<a href="mailto:someone@example.com">someone@example.com</a>
<a href="tel:+496170961709">+49-617-09-617-09</a>
```
I've tried these and they all work correctly on Android, so the correct Intent (browser, email, call) fire by clicking on these Html links.

So currently for me it looks like that it would be better to handle this aspect on the server, just like with any other pre-processing, if they can be done on the server platform independently, that's better. More responsive client with less logic, independently updatable server.

@rutgerg @dmfs @aramvisser 
So what do you think about making all event descriptions html on the server and linkify them, meaning putting plain text fragments that are links in `<a href..>` tags?

dmfs:
Converting a description to HTML is not just a matter of wrapping it in `<p></p>` tags. If you do that, all white spaces (including new lines) will be collapsed to a single white space. Effectively removing any structure from the document.

Just check out this example:
```
<p>
---these
are
five
separate
lines---
</p>
```
renders to

<p>
---these
are
five
separate
lines---
</p>

Also, we should be aware that not all clients may be capable of showing HTML easily. For instance you don't want a voice user interface to spell out "< … p … >" or URLs.

What we really need is both, a rich text and a plain text representation. So the client can choose whichever is more appropriate.

lemonboston:
Oh, right, whitespaces are removed when putting it into `<p>`.
Searched for converter, found this: http://stackoverflow.com/a/21610526/4247460

By _rich text_ you mean _html_, or something else?
I see that providing both plain and html could potentially be useful, maybe it could be a request header to specify which one to return, to avoid bigger payload.
And if there is an html, converting it to plain text is the easier direction I suppose (on the server).

dmfs:
Rich text can be HTML, but could be other formats as well.

I'll bring this topic up in CalConnect TC-API, because I'd prefer a standardized way to deliver rich text. At present we actually don't comply with the draft because we send HTML when the current spec only supports plain text.

lemonboston:
@dmfs Should I continue with this PR then, so that we have some improved linkification handling for plain text in the client for now?

dmfs:
yes, please continue with this, because we won't see the changes in the API any time soon

lemonboston:
@dmfs Ready for review.

dmfs:
@lemonboston please resolve the conflicts

lemonboston:
Conflicts resolved.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/linkify/CompleteWebUrlLinkifying.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.regex.Matcher;
+
+
+/**
+ * A {@link Spannable} that linkifies web urls that starts with "http" or "Http" in the input {@link Spannable}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class CompleteWebUrlLinkifying extends AbstractSpannable
```
dmfs: The name is not correct. It should be "Linkified" instead of "Linkifying", because it doesn't decorate behavior but properties, i.e. it's a "linkified Spannable". Maybe call it `HttpUrlLinkified`

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/linkify/EmailLinkifying.java
```diff
@@ -15,39 +15,33 @@
...
 
 
 /**
- * Linkified text.
+ * A {@link Spannable} that linkifies email addresses in the input {@link Spannable}.
  *
  * @author Gabor Keszthelyi
  */
-public final class Linkified extends AbstractSpanned
+public final class EmailLinkifying extends AbstractSpannable
```
dmfs: Same here: `EmailLinkified`

dmfs (changes requested):

lemonboston:
I've changed the name of both decorators, ready for check again.

dmfs (approved):

PR #248, opened 2017-05-04
PR for 228: Fix font weights on details screen

### LINKED ISSUE #228: fix font weight of action texts
According to the design, the action text elements (including venue and directions) should use Roboto Medium instead of Roboto Regular like they currently do.

lemonboston:
Fixes #228


dmfs (approved):
Looks good.

Issue #212, opened 2017-04-18
Guarantee to not show empty event description


lemonboston:
I saw an event with html description `<p></p>`, which results in an empty text after formatting.
The reason why the description view was correctly not shown in this case is that the `ExpandableTextView` that we use sets itself to `GONE` when it receives an empty text. But we should not rely on this but instead update the logic to hide the view when the formatted text results in an empty one.

PR #249, opened 2017-05-04
PR for 212: Guarantee to not show empty event description

lemonboston:
Implements #212.


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/EventDescriptionText.java
```diff
@@ -0,0 +1,63 @@
...
+            @Override
+            public Optional<CharSequence> create()
+            {
+                if (!description.isPresent())
+                {
+                    return Absent.absent();
+                }
+
+                String desc = description.value();
+                CharSequence formatted = isHtml(desc) ? new Html(desc) : new Linkified(new Trimmed(desc));
```
dmfs: This functionality should be split into two decorators `Styled` and `NonEmpty`, with `NonEmpty` being absent if the delegate is empty.
`EventDescriptionText` is way to specific and also violates SRP.

So this
```
EventDescriptionText formattedDescription = new EventDescriptionText(description);
```
should read like this:
```
Optional<CharSequence> formattedDescription = new NonEmpty(new Styled(description));
```

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/EventDescriptionText.java
lemonboston: Absolutely right, thanks, I've updated it, looks better this way. I've also added `OptionalMapped` that maps from optional to optional - not sure if that's a good name.
Also, I've noticed an embarrassing bug in `Trimmed`, fixed that, added unit test.
Please check again.

dmfs (approved):

Issue #115, opened 2017-02-02
Specify module in gradle dependency exclude to try fixing jitpack import


lemonboston:
In order to not force me to use newer library versions, as an SDK integrator I want to be able to import the event-discovery SDK via jitpack with the library dependency versions as defined by the SDK.
---

In our current setup jitpack doesn't honor the `exclude` statement for the `flexibleadapter` library, possibly because module is not specified, only group, which might be required as per
https://github.com/dcendents/android-maven-gradle-plugin#usage
(This results in effective 25.1.0 support lib version, rather than 25.0.0 as in our demo app.)

lemonboston:
Done it in https://github.com/schedjoules/android-event-discovery-sdk/pull/325 .

PR #251, opened 2017-05-04
PR for 243: Upgrade Android support libraries to latest version

lemonboston:
Implements #243.

@dmfs I've added gradle variables for `compileSdkVersion, buildToolsVersion, minSdkVersion, targetSdkVersion` as well, although I remember you had some concern about that, but I forgot. I can remove them of course. I thought that if `HTTP_CLIENT_ESSENTIALS_VERSION` works without problem than those should be okay as well.

### DIFF
--- a/eventdiscovery-demo/build.gradle
+++ b/eventdiscovery-demo/build.gradle
@@ -28,12 +28,12 @@ def gitCommitNo = { ->
 }
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "25.0.0"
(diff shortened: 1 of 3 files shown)


lemonboston:
As discussed, changed to use 25.0.1 support lib version, which was the effective one, instead of the latest 25.3.1. Rebased.

Note that we are still not enforcing the version with this, just matching with the effective one, in order to enforce we need to do https://github.com/schedjoules/android-event-discovery-sdk/issues/115.

dmfs (approved):

Issue #225, opened 2017-04-25
Align text elements on details page


dmfs:
All texts with an icon in front should be aligned to the key line at 72dp. This also applies to the collapsed title and date in the header.
The description should be horizontally aligned to the icons, which means it needs an inset of 2dp to compensate the icon padding (for symmetry that inset should be applied on the left and on the right side).
Actions should be right aligned with the description (i.e. have the same inset on the right).

PR #253, opened 2017-05-05
Adjust text alignments on the details screen. #225

lemonboston:
Implements #225.

@dmfs I've only applied the 2dp inset for the text, haven't moved icon texts to 72dp keyline because that looks too much for me. Could you please check it for yourself?
Currently it's 16dp padding + 24dp icon + 16dp distance = 56dp
The distance is the `schedjoules_event_details_item_icon_and_text_distance` dimen, so if that is changed to 32dp, we get to the 72dp which looks like this:
device-2017-05-05-171245

I think the fact that we have the full width description changes the situation a bit. Also we may need the real estate for longer venue names and directions, and for the titles as well.
The current state looks more in line with InVision design as well.


wilfredstegeman:
I suggest to keep it as is.

dmfs:
Ok, @lemonboston please just make sure the action items are aligned with the text elements on the collapsed title and the expanded title is aligned with the description.

lemonboston:
@dmfs 
I couldn't find how I could change the padding/margin of the collapsed title, so I increased the action text margins instead with 4dp. Also, the expanded title gets aligned with 17dp strangely (not 18dp as it should be).
I know these are not standard grid values, but this is how I could do it in about an hour and a half, please check if this is good or not, should I spend more time on it.

If collapsed title should come left instead, do you have a tip where to set that?
(I don't even see the title TextView in the view hiearchy snapshot, I suppose it's something drawn..)

eventdiscovery-sdk/src/main/res/layout/schedjoules_details_header.xml
```diff
@@ -16,8 +16,8 @@
                 android:fitsSystemWindows="true"
                 app:expandedTitleGravity="bottom|left"
                 app:expandedTitleTextAppearance="@style/SchedJoules_ExpandedTitle"
-                app:expandedTitleMarginStart="@dimen/schedjoules_activity_horizontal_margin"
-                app:expandedTitleMarginEnd="@dimen/schedjoules_activity_horizontal_margin"
+                app:expandedTitleMarginStart="17dp"
```
dmfs, line 19: I don't understand this. In your comment you state that the expanded title strangely gets aligned to 17dp instead of 18dp, but you explicitly set the margin to 17dp?! Am I missing something?

lemonboston: Sorry, I didn't express it well, I meant that it gets aligned correctly with the other texts when its margin is set to 17dp.

eventdiscovery-sdk/src/main/res/layout/schedjoules_view_event_details_description.xml
```diff
@@ -7,7 +7,7 @@
             android:layout_height="wrap_content"
             android:layout_width="match_parent"
             android:paddingLeft="@dimen/schedjoules_activity_horizontal_margin"
```
dmfs: Shouldn't the description have a left padding of `schedjoules_activity_horizontal_margin_with_icon_padding_inset` to align it with the icon padding instead of the icon boundaries?

eventdiscovery-sdk/src/main/res/layout/schedjoules_details_header.xml
dmfs: ah, strange indeed. Yes the title is drawn. Looks like the missing 1dp is added in code.

dmfs (changes requested):
Apart from the description padding it looks ok to me.

eventdiscovery-sdk/src/main/res/layout/schedjoules_view_event_details_description.xml
lemonboston: Oh, good catch! Yes, that was the whole point for that dimen, I accidentally changed it back in one of my last changes (when I changed it for the action and direction views, cause that don't need the left padding to change)
I've fixed it.

dmfs (approved):

Issue #221, opened 2017-04-24
Show sale status information on ticket button


dmfs:
In order to have my expectations manged properly, as a user I don't want to see a ticket button if there are no tickets available.

lemonboston:
I missed what was this exactly about, could someone remind me please?
Currently we don't show the ticket button if there is no `http://schedjoules.com/rel/action/book` action link with `http://schedjoules.com/booking/type` as `ticket`. (If button was shown when this was not present, then that's a bug.)
Or is this about expired tickets? Is there a parameter for that that we should use?
Thanks.

lemonboston:
@aramvisser Could you please remind us about this? What is that case when there is _book_ action link with _ticket_ subtype, but in fact no ticket available for the event - what is the extra parameter we need to check for?

rutgerg:
There is a sale-status attribute in actions. It currently has 6 options : unknown, available, unavailable, unlimited, multiple, donation. You are looking for unavailable I guess.

A better option imo than to remove the button is to show a button that says 'no tickets'. That could still be clickable.

lemonboston:
Thanks, I see the `sale-status` attribute:

```json
 "links": [
     {
         "rel": "http:\/\/schedjoules.com\/rel\/action\/book",
         "href": "https:\/\/actions.schedjoules.com\/actions\/85cf7098d8e97543",
         "properties": {
             "http:\/\/schedjoules.com\/props\/booking\/type": "ticket",
             "http:\/\/schedjoules.com\/props\/booking\/price-min": 8.51,
             "http:\/\/schedjoules.com\/props\/booking\/price-max": 8.51,
             "http:\/\/schedjoules.com\/props\/booking\/currency": "EUR",
             "http:\/\/schedjoules.com\/props\/booking\/sale-start": "2017-02-03T10:00:00Z",
             "http:\/\/schedjoules.com\/props\/booking\/sale-end": "2017-05-06T06:00:00Z",
             "http:\/\/schedjoules.com\/props\/booking\/sale-status": "unavailable",
             "http:\/\/schedjoules.com\/props\/vendor\/icon": "data:image\/png;base64,XXXXXXXXX"
         }
     },
     {
         "rel": "http:\/\/schedjoules.com\/rel\/action\/add-to-calendar"
     },
     {
         "rel": "http:\/\/schedjoules.com\/rel\/action\/share",
[...]

PR #260, opened 2017-05-15
Change Ticket button label and appearance when sale-status is 'unavai…

lemonboston:
…lable'. #221 

---

TODOs:
* backend: all events now seems to have ticket with sale-status "unavailable"
* finalize color and label for unavailable ticket button
* translations for the new label


lemonboston:
Added the translations, made the dark theme ticket background darker, squashed and rebased.
Ready for review.

Please note that all events I've checked seem to have 'unavailable' status now, so I guess it makes sense to wait with merging until we see if it makes sense to use this.
(The branch `mx/245-simple-response-logging` is rebased on this one now, so you can use that to check the response logs while using the app.)

eventdiscovery-sdk/src/main/res/values/strings.xml
```diff
@@ -30,6 +30,7 @@
 
     <!-- Label for the tickets button on the Event Details screen. -->
     <string name="schedjoules_event_details_ticket_button_title">Get Tickets</string>
+    <string name="schedjoules_event_details_ticket_button_title_not_available">No tickets available</string>
```
dmfs, line 33: Please add a comment that mentions the purpose of this string.

eventdiscovery-sdk/src/main/res/values/theme_dark.xml
```diff
@@ -27,6 +27,7 @@
         <item name="schedjoules_cardBackgroundColor">@color/schedjoules_background_panel_dark</item>
         <item name="schedjoules_googleAttribution">@drawable/powered_by_google_dark</item>
         <item name="schedjoules_schedJoulesAttribution">@drawable/schedjoules_footer_powered_by_dark</item>
+        <item name="schedjoules_disabledButtonColor">@color/schedjoules_dark_disabled_button_background</item>
```
dmfs, line 30: The name is misleading. The button is not really disabled. It's just less prominent. Something like "reluctant" might be a better description. Alternatively we could try a regular non-colored button style (so the background of the butotn in the white theme is white as well). Maybe that's a better idea.

eventdiscovery-sdk/src/main/res/values/colors_light.xml
```diff
@@ -15,6 +15,8 @@
     <color name="schedjoules_empty_thumbnail_background">#dbdbdb</color>
     <color name="schedjoules_empty_thumbnail_draw">#bababa</color>
 
+    <color name="schedjoules_disabled_button_background">#bdbdbd</color>
```
dmfs, line 18: Same here, the name is misleading.

eventdiscovery-sdk/src/main/res/values/colors.xml
```diff
@@ -4,4 +4,5 @@
 
     <color name="schedjoules_green_haze">#0CAF60</color>
     <color name="schedjoules_white">#ffffff</color>
+
```
dmfs, line 7: blank line?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/BasicTicketInfo.java
```diff
@@ -0,0 +1,52 @@
...
+
+
+    public BasicTicketInfo(Link ticketLink)
+    {
+        mLink = ticketLink;
+    }
+
+
+    @Override
+    public boolean isAvailable()
```
dmfs, line 48: In cases like this I'm really annoyed by the design of Java's `Boolean` class. If it was an interface we wouldn't need this kind of class and method. We could just do

```
Boolean available = new TicketAvailability(ticketLink);
```

But that doesn't work unfortunately (probably because it wouldn't work with auto-unboxing).

However, how about something like this instead?
```
boolean available = new TicketAvailability(ticketLink).get();
```
with `TicketAvailability` being a `Lazy<Boolean>`?

dmfs (changes requested):
done with the first round. We should indeed wait for the "unavailable" issue to be resolved before we merge this.

dmfs:
Until the unavailable issue has been resolved you don't need to spend any time with this. We have other important things in the queue.

Issue #262, opened 2017-05-16
Split EventListFragment into a Toolbar and a List fragment


lemonboston:
In order to be able to place the categories loading into the event list loading flow, as a developer, I want to have a more modular structure of the event list screen components
---

PR #264, opened 2017-05-16
Move event list items loading and displaying related code into new ch…

lemonboston:
…ild fragment. #262 

---

I've moved the actual list related code into a new `EventListListFragment` which is used as a child fragment of `EventListFragment`.
`EventListFragment` still contains all the Toolbar stuff, I haven't separated that to a fragment because it's not need at this stage and maybe not needed later either. There will be some things to figure out/decide about how to show the categories filter on the very first load. So I just want to defer any decision to where we actually have to make it.

Next step is to make the current `EventListListFragment` a `MicroFragmentHost` and create its loader, error, and show MFs.

The change in this PR - which went surprisingly without issues - already reduces complexity, for example the former `onResume()` logic disappeared (!). So I think this could be merged to master instead of the categories-master.


dmfs (approved):
Works for me.

PR #268, opened 2017-05-17
Upgrade MicroFragments to latest version. Implements #266

dmfs:
With the new version MicroFragments no longer have to add the FragmentEnvironment themselves to the Fragment arguments.
Also, the back button is handled by the hosting activity now and not by the HostFragment.


lemonboston (approved):
Looks good.

PR #272, opened 2017-05-18
Restructuring towards single Activity + MicroFragments architecture (…

lemonboston:
…by eliminating EventListActivity, FeedbackActivity). #72 

---

Ready for review.
I've left a few TODOs here and there to keep the changeset from growing too big.


lemonboston:
Known issue is as you mentioned that the list reloads now after coming back from details screen.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/view/EventListMenu.java
```diff
@@ -60,6 +62,8 @@ public boolean onOptionsItemSelected(MenuItem item)
         }
         else if (item.getItemId() == R.id.schedjoules_event_list_menu_feedback)
         {
+            // TODO Needed to close the menu explicitly, otherwise the popup was there when came back from Feedback. Why is this, remove if possible.
```
dmfs: check out how it's done in the event details MF. IIRC you need to "post" the handler so it's enqueued in the message queue.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListMicroFragment.java
```diff
@@ -0,0 +1,109 @@
...
+    {
+        mArgs = args;
+    }
+
+
+    @NonNull
+    @Override
+    public String title(@NonNull Context context)
+    {
+        throw new UnsupportedOperationException("Title is not used for this fragment.");
```
dmfs, line 53: We probably should remove that method from the MF framework.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/UidEventDetails.java
```diff
@@ -46,10 +49,11 @@ public UidEventDetails(@NonNull String eventUid)
     @Override
     public void show(@NonNull Activity activity)
     {
-        Intent intent = new Intent(activity, MicroFragmentHostActivity.class);
-        Bundle nestedBundle = new Bundle(1);
-        nestedBundle.putParcelable("MicroFragment", new EventLoaderMicroFragment(mEventUid));
-        intent.putExtra("com.schedjoules.nestedExtras", nestedBundle);
-        activity.startActivity(intent);
+        activity.startActivity(
+                // TODO Use action intent instead?
```
dmfs: Why would we want to use an action intent?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
```diff
@@ -55,54 +59,43 @@ public BasicEventDetails(@NonNull Event event)
...
-                if (actionService.isConnected())
+                try
                 {
-                    actionService.disconnect();
+                    return result.value();
+                }
+                catch (Exception e)
+                {
+                    // Should not happen
+                    throw new RuntimeException(e);
```
dmfs: Why are you using a `SafeAsyncTask` if that should never happen? Wouldn't it make more sense to use a regular `AsyncTask` if the background operation never throws?

dmfs (changes requested):
First round done.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/view/EventListMenu.java
lemonboston: I see, I've updated it to use the same.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListMicroFragment.java
lemonboston: Agree, created an issue for it https://github.com/dmfs/MicroFragments/issues/3.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/UidEventDetails.java
lemonboston: Not sure, but wasn't it the general suggested approach to use actions when possible? To keep things a bit more decoupled maybe. To be able to add a proxy activity for example if needed.
So the question is general in this sense. But I don't see any concrete need here.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
lemonboston: I used it for the `WeakReference` callback. I am generally a bit concernced about all these `AsyncTasks` by the way, but that's another topic to figure out something better.
So I can use a regular `AsyncTask` here or create one with only the `WeakReference` callback.

But a related thing here, can we fire the transaction from a background thread? There's a TODO about that there.
If we can, I would probably make this current `LoadActionsFromCacheTask` just `Callable<Optional<List<Link>>>`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/UidEventDetails.java
dmfs: No, we don't use Actions "when possible" (I guess it would always be possible). But we use Action Intents to dispatch activity launches (instead of doing that in our code), whenever possible.

Consider the add-to-calendar intent. Instead of checking if the hosting app provides its own activity to handle that and call the default "insert event intent" otherwise, we always fire the add-to-calendar intent and let the system figure out if there is an intent-filter with a higher priority than our default handler.

In this case, there is no need for the hosting app to intercept that call. It's a purely internal thing and there is nothing to dispatch.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
dmfs: Yes, you can fire transactions from a background thread. The ParcelableDovecote (which is used internally by the MF framework) makes sure that the pigeon is always delivered on the main thread: https://github.com/dmfs/PigeonPost/blob/master/localbroadcast-pigeon/src/main/java/org/dmfs/pigeonpost/localbroadcast/ParcelableDovecote.java#L97

dmfs: In this particular case it's certainly enough to just launch a Thread (like before). The thread won't run much longer than ~40 milliseconds, so there is little risk in leaking any context.

dmfs: Btw, you can create a Timestamp before you start the Thread and pass that one to the `ForwardTransition`. That's actually the best way to do it. There is no need to create the `ForwardTransition` on the UI thread. It's sufficient if you pass a UiTimeStamp.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/UidEventDetails.java
lemonboston: Ok, thanks for expanding on this a bit.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
lemonboston: Ok, I've updated it.

lemonboston:
Ready for checking again.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
```diff
@@ -55,54 +63,48 @@ public BasicEventDetails(@NonNull Event event)
...
-     */
-    private final static class StartEventDetailsRunnable implements Runnable
-    {
-        private final Activity mActivity;
-        private final Event mEvent;
-
+        final OptionalArgument<MicroFragmentHost> host = new OptionalArgument<>(Keys.MICRO_FRAGMENT_HOST, activity);
 
-        private StartEventDetailsRunnable(@NonNull Activity activity, @NonNull Event event)
+        if (!host.isPresent())
```
dmfs: Why is it an `OptionalArgument` if we fail anyway if it's not present?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
```diff
@@ -55,54 +63,48 @@ public BasicEventDetails(@NonNull Event event)
...
-
-    /**
-     * A {@link Runnable} that starts the details activity in the background after trying to load any actions from the cache.
-     */
-    private final static class StartEventDetailsRunnable implements Runnable
-    {
-        private final Activity mActivity;
-        private final Event mEvent;
-
+        final OptionalArgument<MicroFragmentHost> host = new OptionalArgument<>(Keys.MICRO_FRAGMENT_HOST, activity);
```
dmfs: The type of the variable should be `Optional<MicroFragmentHost>`

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
```diff
@@ -55,54 +63,48 @@ public BasicEventDetails(@NonNull Event event)
...
             }
-            finally
+
+
+            private Optional<List<Link>> loadActionsFromCache(FutureServiceConnection<ActionService> actionService)
             {
-                if (actionService.isConnected())
+                try
+                {
+                    List<Link> actions = actionService.service(40).cachedActions(mEvent.uid());
```
dmfs: please get rid of the temp variable. You can easily put this into a one-liner without reducing readability.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/activities/MicroFragmentHostActivity.java
```diff
@@ -46,6 +50,8 @@
     protected void onCreate(Bundle savedInstanceState)
     {
         super.onCreate(savedInstanceState);
+
+        // TODO No need to set layout, android.R.id.content can be used for the fragment containerId
```
dmfs: Is it safe to do so? If so, why not do it right away?

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
lemonboston: It was for being able to signal that `BasicEventDetails` can only be used from Activtiy with MFHost. But you are right, it's unnecessary, so I removed it and use it as mandatory with new convenience class `ActivityMicroFragmentHost implements Lazy<MicroFragmentHosts>`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
lemonboston: This was removed.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/eventdetails/BasicEventDetails.java
lemonboston: Done.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/activities/MicroFragmentHostActivity.java
lemonboston: I think it's safe to do, they say it here as well:
https://stackoverflow.com/questions/24712227/android-r-id-content-as-container-for-fragment

So I changed it now, for `LocationPickerActivity` as well, and removed the layout.

lemonboston:
I hit a crash accidentally with "Don't keep Activities" turned on, when returning from `LocationPickerActivity`: the host is not in the `Intent`, even thought it was set with `setIntet()`. I am looking into this now.

lemonboston:
Ah, we can't use the `Intent` for saving state:
http://blog.jjyao.me/blog/2014/08/28/which-intent-will-you-get-after-android-relaunches-the-activity/
(nice post)

> Because you don’t know which intent you will get after Android relaunches the activity, you should use savedInstanceState to restore activity’s state. In fact, ActivityRecord stores activity’s savedInstanceState in its icicle variable, so it won’t be lost!

lemonboston:
Fixed it by using `savedInstanceState` as well. We could probably start to do that single state thing here you thought about, as a next task to clear this up a bit. Going to chat you on slack..

dmfs:
Too bad. So you have to revert to storing it in saved instance state like before.

lemonboston:
Ready for review again.

dmfs (approved):

lemonboston:
I vote for 13 story points for this (feel free to set any).

dmfs:
adding 13 to #72

Issue #261, opened 2017-05-15
Link property key mismatch with server


lemonboston:
Some of the action link parameter key uris hardcoded in the client are not matching with the ones used by the server.
e.g.: `ParcelableLink` uses `http://schedjoules.com/booking/type`, while it is `http://schedjoules.com/props/booking/type` in the response.

API documentation:
https://schedjoules.github.io/event-discovery-api/#get-event-actions

Double-check all other hardcoded values agains the API docs.

Optionally create unit test for parsing the sample json from the API docs if it makes sense.

PR #274, opened 2017-05-18
Fix link key mismatches between client and API. Fix bug about incorre…

lemonboston:
…ct booking type check. Move all link constants to one place. #261 

---

In addition to double-checking all Link string literals against the API documentation and also checking the rest in logs, I've also moved all of them as constants to one class. I think it gives a better overview, easier to double-check them later as well, less chance for mistake by defining them scattered in the codebase. Static inner classes are used so they aren't even initialized until used. That can be further refined if needed.
The property type constants are defined as `PropertyType` already for convenience.

I've also fixed the bug in this pull request as well about the incorrect check for the booking type.
(It is also fixed in another PR that may not make master eventually (about ticket statuses)).


dmfs:
I just thought that this would make more sense to be in the `java-api-client`. Is there anything Android specific in these rel types?

lemonboston:
I thought about that, I think that the only reason for keeping it in the SDK is that if any of it changes, for example we add new ones, we don't need another `java-api-client` release. It's a bit similar to what you said about the benefit of the JSON format in general, so it can be changed without changing 'intermediaries', validators, etc.
Also, not all properties are defined now I think.. for example sale-start, sale-end, ticket-price.

dmfs (approved):
I general I think this should go to the java-api-client. But it's certainly better to wait for the rel-types to stabilize a bit more. Let's get back to this in a couple of months.

lemonboston:
I vote for 3 story points for this.

Issue #273, opened 2017-05-18
Don't reload event list when coming back from event details


lemonboston:
This is a known intermediary issue during the refactoring that prepares the structures for the categories.
It is separated to a task to avoid other pull request growing too big.

lemonboston:
Fixed on branch `categories/273-dont-reload-list-from-details`, will submit PR after #72 is merged to `categories/master`.

lemonboston:
The lifecycles we talked about:

Normal Fragment:
1.. First initialization
onCreate(), onCreateView(), savedState=null
2.. Rotation
new Fragment instance, onCreate(), onCreateView(), savedState!=null
3.. Popped from backstack
same Fragment instance, onCreateView(), savedState=null

Retained Fragment (1 and 3 are the same):
2.. Rotation
same Fragment instance, onCreateView(), savedState!=null

I will update the code shortly with what I have in mind to handle case 3.

lemonboston:
When normal fragment is on backstack while rotating the screen:
immediately recreated: new Fragment instance, onCreate(), savedState!=null
when popping back from back stack: same Fragment instance, onCreateView(), savedState!=null

Logs:
```
05-19 14:47:48.402 D/Investigator: [main] EventDetailFragment{e41335f #2 id=0x7f0d0009}.onCreate() | savedState = null
05-19 14:47:48.403 D/Investigator: [main] EventDetailFragment{e41335f #2 id=0x7f0d0009}.onCreateView() | savedState = null

// Rotation while note having EventDetailFragment on screen (on Feedback)
05-19 14:48:12.807 D/Investigator: [main] EventListFragment{7a05c9c #0 id=0x7f0d0009}.onCreate() | savedState = Bundle[{android:support:fragments=android.suppo....
05-19 14:48:12.815 D/Investigator: [main] EventDetailFragment{321c607 #2 id=0x7f0d0009}.onCreate() | savedState = Bundle[{android:suppo....

// Back to event details
05-19 14:48:27.670 D/Investigator: [main] EventDetailFragment{321c607 #2 id=0x7f0d0009}.onCreateView() | savedState = Bundle[{android:support:fragments=android.sup.....
```

lemonboston:
I vote for 5 story points for this, because of the effort involved to investigate, log and collect those lifeycycle scenarios and the discussion.
Could be 3 as well if you see that more fit.

dmfs:
ok, let's go with 5

PR #275, opened 2017-05-19
Fix event list reloading when coming back from event details. #273

lemonboston:
---

This will need to be rebased on `categories/master` first after #72 is closed.

This fix is only applicable now to retained fragments, if we encounter any other problem with normal fragments popped from back stack, we need to revisit the general topic. @dmfs the spreadsheet I shared with you contains I think a good overview of the lifecycles, so we can look into that again, if needed.

I've also changed the UP navigation in the Details MF to fire the BackTransition as you said.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/EventDetailFragment.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/EventDetailFragment.java
@@ -40,6 +40,8 @@ import com.schedjoules.eventdiscovery.framework.microfragments.eventdetails.frag
 import com.schedjoules.eventdiscovery.framework.utils.InsightsTask;
 
 import org.dmfs.android.microfragments.FragmentEnvironment;
+import org.dmfs.android.microfragments.MicroFragmentEnvironment;
+import org.dmfs.android.microfragments.transitions.BackTransition;
 import org.dmfs.httpessentials.types.StringToken;
 
 
@@ -56,7 +58,8 @@ public final class EventDetailFragment extends BaseFragment implements EventDeta
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
     {
-        ShowEventMicroFragment.EventParams parameters = new FragmentEnvironment<ShowEventMicroFragment.EventParams>(this).microFragment().parameter();
+        final MicroFragmentEnvironment<ShowEventMicroFragment.EventParams> environment = new FragmentEnvironment<>(this);
+        ShowEventMicroFragment.EventParams parameters = environment.microFragment().parameter();
 
         if (savedInstanceState == null)
         {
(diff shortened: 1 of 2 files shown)


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/EventDetailFragment.java
```diff
@@ -56,7 +57,8 @@
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
     {
-        ShowEventMicroFragment.EventParams parameters = new FragmentEnvironment<ShowEventMicroFragment.EventParams>(this).microFragment().parameter();
+        final FragmentEnvironment<ShowEventMicroFragment.EventParams> environment = new FragmentEnvironment<>(this);
```
dmfs: Use `MicroFragmentEnvironment` for the variable type.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/common/BaseBackStackedRetainedFragment.java
```diff
@@ -0,0 +1,59 @@
...
+        super.onCreate(savedInstanceState);
+        setRetainInstance(true);
+    }
+
+
+    @Nullable
+    @Override
+    public final View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState)
+    {
+        View view = onCreateView(inflater, container, savedInstanceState, mIsInitializing);
```
dmfs: I'm not a big fan of the template method pattern.
Also, I consider this only a temporary workaround due to the lack of a better backstack implementation. No other Fragment we create should need this, so think it's not worth creating a separate abstraction class for this. That only encourages using this more often when we really should search for a better solution instead.

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/EventDetailFragment.java
lemonboston: Done.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/common/BaseBackStackedRetainedFragment.java
lemonboston: Ok, I removed the base class and use the logic in `EventListListFragment` instead.

lemonboston:
Changes done, but I suggest we wait until #272 is merged, then I squash and rebase on `categories/master` so you can have a final look.

dmfs:
please resolve conflicts and rebase

lemonboston:
Resolved and rebased.

dmfs (approved):
Looks good and appears to work well.

PR #277, opened 2017-05-22
Update java-api-client to latest version and make Event Durations opt…

### LINKED ISSUE #196: Hide end time in UI
The UI currently shows the default duration of 2 hours for every event. However, for pretty much all events this is just a default (not even an estimate) and can be very wrong. In this case it's probably better to give no information than to give wrong information. So the end time should not be shown until we have a better mechanism to distinguish actually known durations, estimates and defaults.

dmfs:
…ional. Fixes #196.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/AddToCalendarActionExecutable.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/AddToCalendarActionExecutable.java
@@ -30,6 +30,7 @@ import com.schedjoules.eventdiscovery.framework.EventIntents;
 import com.schedjoules.eventdiscovery.framework.utils.InsightsTask;
 
 import org.dmfs.httpessentials.types.Link;
+import org.dmfs.rfc5545.Duration;
 
 import static com.schedjoules.eventdiscovery.framework.utils.LocationFormatter.longLocationFormat;
 
@@ -72,7 +73,7 @@ public final class AddToCalendarActionExecutable implements ActionExecutable
         Bundle bundle = new Bundle(8);
         bundle.putLong(CalendarContract.EXTRA_EVENT_BEGIN_TIME, mEvent.start().getTimestamp());
         bundle.putLong(CalendarContract.EXTRA_EVENT_END_TIME,
-                mEvent.start().addDuration(mEvent.duration()).getTimestamp());
+                mEvent.start().addDuration(mEvent.duration().value(mEvent.start().isAllDay() ? new Duration(1, 1, 0) : Duration.parse("PT2H"))).getTimestamp());
         bundle.putBoolean(CalendarContract.EXTRA_EVENT_ALL_DAY, mEvent.start().isAllDay());
         bundle.putString(CalendarContract.Events.TITLE, mEvent.title());
         bundle.putString(CalendarContract.Events.DESCRIPTION, mEvent.description());
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/datetime/DateTimeRange.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/datetime/DateTimeRange.java
@@ -23,6 +23,7 @@ import android.text.format.DateUtils;
 
 import com.schedjoules.client.eventsdiscovery.Event;
 
+import org.dmfs.optional.Optional;
 import org.dmfs.rfc5545.DateTime;
 import org.dmfs.rfc5545.Duration;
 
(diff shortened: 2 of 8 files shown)


dmfs:
Note to self: Fix story number before merging: Should be 195 instead of 196.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/AddToCalendarActionExecutable.java
```diff
@@ -72,7 +73,7 @@ private Bundle extrasForAddToCalendar()
         Bundle bundle = new Bundle(8);
         bundle.putLong(CalendarContract.EXTRA_EVENT_BEGIN_TIME, mEvent.start().getTimestamp());
         bundle.putLong(CalendarContract.EXTRA_EVENT_END_TIME,
-                mEvent.start().addDuration(mEvent.duration()).getTimestamp());
+                mEvent.start().addDuration(mEvent.duration().value(mEvent.start().isAllDay() ? new Duration(1, 1, 0) : Duration.parse("PT2H"))).getTimestamp());
```
lemonboston, line 76: Could we have this as something like `FallbackDuration implements Lazy<Duration>`?:
```
mEvent.start().isAllDay() ? new Duration(1, 1, 0) : Duration.parse("PT2H")
```
So this fallback value definition is not that hidden and mingled with the calendar event setup.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/datetime/DateTimeRange.java
```diff
@@ -35,10 +36,10 @@
 public final class DateTimeRange implements FormattedDateTime
 {
     private final DateTime mStartTime;
-    private final Duration mDuration;
+    private final Optional<Duration> mDuration;
 
 
-    public DateTimeRange(DateTime startTime, Duration duration)
+    public DateTimeRange(DateTime startTime, Optional<Duration> duration)
```
lemonboston, line 42: There is `mDuration == null` check further below, could be removed now.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/EventHeaderDateTimeView.java
```diff
@@ -46,6 +46,8 @@ public EventHeaderDateTimeView(SchedjoulesDetailsHeaderBinding binding)
     @Override
     public void update(Event event)
     {
+        // TODO: go back to this version once the API stops to return "estimated" end times:
+        // CharSequence rangeCharSequence = new DateTimeRange(event).value(mDateTimeViewCollapsed.getContext());
```
lemonboston, line 50: The 'interoperability' of `DateTimeRange` and `LongDateAndTimeNoYear` is not clear for me here, do they surely only differ in showing end time as well. But it's probably not important at this point, I guess we can double check when we switch back to showing the end times as the TODO says.
We could possibly create an `EventDateTime` that delegates based on whether duration is present or not, if we need to use 2 formatters.

lemonboston (approved):
Overall looks good for me.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/EventHeaderDateTimeView.java
dmfs: This is not really about the `Duration` being present or not. `DateTimeRange` actually handles absent and present `Duration`s.
At present the API always returns a default duration of 2 hours, but that's wrong in most cases. So we currently use `LongDateAndTimeNoYear` which doesn't show an end time at all (even if the duration is present). We'll keep this in place until the API stops sending "default" durations of 2 hours. At that time we switch to `DateTimeRange` which only shows an end time if the duration is present.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/datetime/DateTimeRange.java
dmfs: done

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/actions/AddToCalendarActionExecutable.java
dmfs: Maybe later. The default is really just needed when the event is added to the calendar. It shouldn't be necessary anywhere else in our own UI code. So this is really just a workaround. I hope that we'll be able to predict durations more accurately, eventually.

dmfs:
I've removed that null check. Since you already approved the other changes I'll go ahead and merge.

Issue #276, opened 2017-05-19
Create loader splash screen MicroFragment for List Screen


lemonboston:

lemonboston:
@dmfs How should we handle error here? Show a white background, dark text message in middle that "Loading failed, tap to retry" and then load the splash again when tapped?
Should we do that in a separate story?

lemonboston:
A close to ready state on `categories/stories/276-loader-splash-MF-for-List` branch.
Finalizing is blocked until `categories/master` is updated with #272 and #277.

PR #282, opened 2017-05-23
Create splash loading MicroFragment for the list screen. #276

lemonboston:
---

This pull requests creates a loading MF for the List screen. It does not create loader fragment to  `EventListListFragment` yet, so that supports 2 ways of loading here, from initial parameters (needed after new city selected) and from `ResultPage`.

2 known issues:
- A second circle progress bar is shown for a moment after the loading, before showing the list. I don't know what is that, it should not be the one in the list layout. Probably no need to address that before switching to proper splash screen with the logo.
- Bundle is used as MF parameter, I would get back to those at the end of the refactors, to change them maybe to typed ones. In that case we need a that corresponds to EventDiscovery (SDK interface)

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/StructuredResultPage.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/StructuredResultPage.java
@@ -73,7 +73,7 @@ public final class StructuredResultPage<T> implements ResultPage<T>
     @Override
     public ApiQuery<ResultPage<T>> previousPageQuery() throws IllegalStateException
     {
-        if (mPrevPageQuery == null)
+        if (mIsFirstPage)
         {
             throw new IllegalStateException("No previous page query");
         }
@@ -84,7 +84,7 @@ public final class StructuredResultPage<T> implements ResultPage<T>
     @Override
     public ApiQuery<ResultPage<T>> nextPageQuery() throws IllegalStateException
     {
-        if (mNextPageQuery == null)
+        if (mIsLastPage)
         {
             throw new IllegalStateException("No next page query");
         }
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/StructuredResultPage.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/StructuredResultPage.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.model;
+
+import android.support.annotation.Nullable;
+
+import com.schedjoules.client.ApiQuery;
+import com.schedjoules.client.eventsdiscovery.ResultPage;
+
+
+/**
+ * A {@link ResultPage} that takes the ready properties as constructor parameters.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class StructuredResultPage<T> implements ResultPage<T>
+{
+    private final Iterable<T> mItems;
+    private final boolean mIsFirstPage;
+    private final boolean mIsLastPage;
+    private final ApiQuery<ResultPage<T>> mPrevPageQuery;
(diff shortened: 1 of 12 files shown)


lemonboston:
And the most important missing point is error handling as asked in the story.

lemonboston:
I've created separate story for the error handling part #284 .
So this is ready for review from my side.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/StructuredResultPage.java
```diff
@@ -0,0 +1,93 @@
...
+    public boolean isLastPage()
+    {
+        return mIsLastPage;
+    }
+
+
+    @Override
+    public ApiQuery<ResultPage<T>> previousPageQuery() throws IllegalStateException
+    {
+        if (mPrevPageQuery == null)
```
dmfs: This should depend on `isFirstPage` not on the page query being `null` (which would be a bug).
Same goes for the next page.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/EventEnvelopeBox.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags)
+    {
+        dest.writeString(mEnvelope.etag());
+        dest.writeString(mEnvelope.uid());
+        dest.writeInt(mEnvelope.hasPayload() ? 1 : 0);
+        dest.writeParcelable(new EventBox(mEnvelope.payload()), flags);
```
dmfs: What do you parcel when there is no payload?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/EventResultPageBox.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+
+    public static final Creator<EventResultPageBox> CREATOR = new Creator<EventResultPageBox>()
+    {
+        @Override
+        public EventResultPageBox createFromParcel(Parcel in)
+        {
+            List<Box<Envelope<Event>>> envelopeBoxes = new LinkedList<>();
+            in.readList(envelopeBoxes, getClass().getClassLoader());
```
dmfs: have you considered to read and write the envelopes with `createTypedList` and `writeTypedList`?

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/model/StructuredResultPage.java
lemonboston: Ok, updated.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/EventEnvelopeBox.java
lemonboston: Good catch, thanks, that would've led to NullPointer probably. Fixed it.
Can we create a story to api-client to use `Optional` here as well?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/EventResultPageBox.java
lemonboston: No, I haven't even noticed that option, I've changed it now to use that one.

lemonboston:
Read for check again.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/EventResultPageBox.java
```diff
@@ -0,0 +1,127 @@
...
+
+
+    public static final Creator<EventResultPageBox> CREATOR = new Creator<EventResultPageBox>()
+    {
+        @Override
+        public EventResultPageBox createFromParcel(Parcel in)
+        {
+            List<Box<Envelope<Event>>> envelopeBoxes = new LinkedList<>();
+            in.readTypedList(envelopeBoxes, EventEnvelopeBox.CREATOR);
+            List<Envelope<Event>> envelopes = new ArrayList<>();
```
dmfs, line 97: Please make a note to apply a `Mapped` `Iterable` decorator on `envelopeBoxes` instead of converting them eagerly once dmfs/iterators#6 has been implemented. `StructuredResultPage` takes an `Iterable` so this should work well.

dmfs (approved):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/EventResultPageBox.java
lemonboston: Added a TODO comment there and pushed to `categories/master` branch.

Issue #278, opened 2017-05-22
Create Loader and Error Fragment for EventListListFragment


lemonboston:
In order to reduce complexity in current event list handling, separate loading and initial error stages into separate Fragments.
---

lemonboston:
I've finished with this on `categories/stories/278-loader-list-fragment` branch, but #276 has to be merged before I can rebase on `categories/master` and submit the pull request.

Issue #284, opened 2017-05-24
Error handling on Event List loader MicroFragment


lemonboston:

lemonboston:
Until categories loading and UI filter displaying is not in place, what we can simply do here is to forward to the EventListFragment with the original parameters. That will try to load again and handle errors.
This would allow to come to a stabilized version of `categories/master` branch a bit quicker, so it can be merged to `master` before adding the categories.

lemonboston:
If we keep the splash loader screen with logo, an error dialog with a message and 'retry' may be better than a full white screen and center message like on the list now.

PR #286, opened 2017-05-24
Create Loader, Error, Show Fragments replacing EventListListFragment. #278

lemonboston:
Submitting this now, but this has to be rebased after #282 is merged.

Code in the original list fragment got significantly simpler by this separation nicely.

Notes:
I've created a `BoxDovecote` which can use a `Box<T>` as pigeon payload and `T` as callback value, used it at one point, but not needed in the end (because of `LoadResult`), still kept it thinking that it might be useful later.

Created this to represent the loading result that can be sent with the pigeon:
```
interface LoadResult<T> extends Parcelable
{
    boolean isSuccess();
    T result() throws IllegalStateException;
}
```
It's very similar to `Optional` but this felt more appropriate for me here. Having an absent value would not be the same semantically for me as having a failed load. Also this is a `Parcelable` so we can use it with `ParcelableDoveCote`.

I've also added some support for standard Fragment transactions.

And a `FragmentBuilder` for convenience, to avoid the `setArgs()` calls, although this builder is not immutable..   so let me know if you'd rather not have that one.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/loadresult/ErrorResult.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/loadresult/ErrorResult.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2017 SchedJoules
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.schedjoules.eventdiscovery.framework.utils.loadresult;
+
+import android.os.Parcel;
+
+
+/**
+ * Error {@link LoadResult}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class ErrorResult<T> implements LoadResult<T>
+{
+    @Override
+    public boolean isSuccess()
+    {
+        return false;
+    }
+
+
+    @Override
+    public T result() throws IllegalStateException
+    {
+        throw new IllegalStateException("Error result has no result");
+    }
+
+
+    @Override
+    public int describeContents()
+    {
+        return 0;
+    }
+
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags)
+    {
+
+    }
+
+
+    public static final Creator<ErrorResult> CREATOR = new Creator<ErrorResult>()
+    {
+        @Override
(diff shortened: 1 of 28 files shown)


dmfs:
please rebase

lemonboston:
Rebased, ready for review. Please check the comments above as well.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/loadresult/ErrorResult.java
```diff
@@ -0,0 +1,75 @@
...
+
+
+    public ErrorResult(Exception causeException)
+    {
+        mCauseException = causeException;
+    }
+
+
+    @Override
+    public T result() throws Exception
```
dmfs: Since you have a specific Exception you should not throw a generic Exception here. Make it specific as well.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/loadresult/LoadResult.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface LoadResult<T> extends Parcelable
+{
+    /**
+     * Returns the load result.
+     *
+     * @throws Exception
+     *         if the load was not successful
+     */
+    T result() throws Exception;
```
dmfs: Same here of course. Throw a `LoaderResultException`.

dmfs (changes requested):
If I interrupt the internet connection, the initial loader fragment seems to keep spinning forever.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/loadresult/ErrorResult.java
lemonboston: Right, I realized this, too, (this is why I logged in now actually to update that). Done.

lemonboston:
> If I interrupt the internet connection, the initial loader fragment seems to keep spinning forever.

Do you mean the splash loader MF? (`EventListLoaderMicroFragment`) That error handling is a known missing part there at this point (TODO at line 200)
The task for that: https://github.com/schedjoules/android-event-discovery-sdk/issues/284
Please see the comments there.

lemonboston:
@dmfs Any progress on this maybe? You mentioned you had some concern, what were they? Can we maybe create ticket for that to the backlog and get back to them after 1st version of categories are implemented?
Note that although 2 or 3 PRs have been merged to this, this alone still does not represent a complete state after the refactors. (Error handling on splash screen, bottom-up transitions are separate tasks for example.)

dmfs:
There is a glitch when I change the location, see the screen shot.
It happened when I changed the location after rotating the device.

screenshot_1498128195

lemonboston:
I've seen that before and I thought I fixed that. The HeaderFragment is added 2 times, it relates to the fragment restoring lifecycles when rotating and popping from backstack.
I've created a bug: https://github.com/schedjoules/android-event-discovery-sdk/issues/315 so this PR can go ahead without it. (I will work on the bug today)

dmfs (approved):

Issue #279, opened 2017-05-23
shadow on status bar is back in categories branch


dmfs:
The MF refactoring brought back the shadow on the status bar after the first transition to event details. Apparently we need a "Header" smartview in every MF which sets the status bar color accordingly.

PR #288, opened 2017-05-24
Fix shadow on status bar after visiting details page. #279

lemonboston:
I've fixed it by turning the transparent status bar on and off in `onResume()` and `onPause()` of the event details fragments. Not sure if this is the best way but we don't need to touch the other fragments with this.

During the transitions the shadow is still there. Not sure if that can be eliminated easily though.
So an option can be, I suppose, is to use this fix and fine tune it later if needed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.2'
+        classpath 'com.android.tools.build:gradle:2.3.3'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/common/BaseFragment.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/common/BaseFragment.java
@@ -19,6 +19,10 @@ package com.schedjoules.eventdiscovery.framework.common;
 
 import android.support.v4.app.Fragment;
 
+import com.schedjoules.eventdiscovery.R;
+import com.schedjoules.eventdiscovery.framework.utils.colors.AttributeColor;
+import com.schedjoules.eventdiscovery.framework.utils.colors.Transparent;
+
 
 /**
  * Base class for all Fragments in the app.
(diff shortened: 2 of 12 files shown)


dmfs (commented):
instead of trying to fix this entirely in the Fragments which need a transparent statusbar, wouldn't it make sense if every MF sets the statusbar it needs itself?

lemonboston:
That would possibly work as well but then I think it would probably be good to have some mechanism that guarantees that we set it. Maybe `BaseFragment` could apply it like the theme in `BaseActivity`, but still need to put it to args. Custom `MicroFragment` extension?

In my opinion we could apply this fix now and get back to it later. But let me know if I should change it, so every fragment sets the status bar.

dmfs:
Yes, I think we should change that. In this case it's probably not so important to enforce setting the color, because you can easily spot during monkey testing if the statusbar is not set correctly. However, instead of setting the color directly how about two instances like `TransparentStatusBar` and `OpaqueStatusBar` which set the color accordingly on their own?

lemonboston:
Ok, I will apply it for every Fragment. Regarding the two instances, I thought about that, but it is a `SmartView` now which need some data to update with. 
Could be a `SmartView<Void[]>` with `update(Void... voids)` so we can write `update()` but it probably doesn't make much sense. So I can only think of that `Configuration` how we could represent this.
`new TrasparentStatusBarConfig().apply(getActivity())`.

lemonboston:
How about a `CoordinatorLayout` subclass instead that sets the status bar color on `onAttachedToWindow()` and `onDetachedFromWindow()`. That looks like the correct lifecycle method to do that since status bar belongs to the `Window`. So we would just use that in layout xml and no need to touch the Fragments.
What do you think?

lemonboston:
Or `CollapsingToolbarLayout` actually instead of `CoordinatorLayout`, since the status bar covering belongs to that.

lemonboston:
I tried that but doesn't work reliably because the loader MF's ToolbarLayout `onDetachedFromWindow()` is only called after the 2nd MF's `ToolbarLayout` `onAttachedToWindow()`:
```
0,0-0,0 #7f0d00b6 app:id/schedjoules_event_detail_toolbar_layout}.onAttachedToWindow()
0,0-0,0 #7f0d00b6 app:id/schedjoules_event_detail_toolbar_layout}.onAttachedToWindow()
0,0-1080,768 #7f0d00b6 app:id/schedjoules_event_detail_toolbar_layout}.onDetachedFromWindow()
```
I've also tried with `onVisibilityChanged()` but that wasn't reliable either.

So I will simply put the updates to every Fragment's onCreateView().

lemonboston:
I've updated and squashed, since the first version was superseded completely.
This has been added to `BaseFragment`:
```
    protected void setStatusBarCoverEnabled(boolean enabled)
    {
        new StatusBar(getActivity()).update(enabled ?
                Transparent.INSTANCE : new AttributeColor(getContext(), R.attr.colorPrimaryDark));
    }
```
Ready for review.

lemonboston:
Rebased to `categories/master`.

dmfs (approved):
Not perfect yet, but ok for now.

Issue #292, opened 2017-05-30
Find a way to functional test serializations


lemonboston:
In order to ensure correct working of the app in special circumstances, as a developer I want to have a way to functionally test that our serialization implementations work correctly.
---

PR #293, opened 2017-05-30
Add utility to force parcelling to be able to test it. #292

lemonboston:
I've already used this on branch 276 to test the parcelling, so I thought we can add this to the code base, so we have it at hand. This is only a util method that can be called in `BundleBuilder` which will ensure deep parcelling of the parameters.

Note: I've also tried testing with "Don't keep Activities" and use the kill process button in Android Studio, but none of them looked appropriate for this.


dmfs:
This is not production code and should go to the test sources. This is just like debug logs, they should not go into production code not even commented out.
Also I dislike the general approach. If you want to test parcelling, write a proper unit test instead.

Issue #296, opened 2017-05-30
(categories-branch) Original start date is not respected after a new city selected


lemonboston:
Steps to reproduce:
1. Open event discovery for future time
2. Go to location picker, select another city
Expected:
3. Event discovery for the new city starts with the original date
Actual:
4. Event discover starts from today

wilfredstegeman:
This is not a bug imo. Choosing a different location resets previous filters, like a date. Check the Facebook Events app for example, it works exactly the same.

dmfs:
In general I agree. If you select a different city it's quite unlikely that you want to search it on the same day.
That will be different if/when we also support to select a specific venue. In that case it might be annoying if you have to go back to the date every time you change the venue.

lemonboston:
@wilfredstegeman I just tried with Facebook Events app on Android, selected 'This Weekend', changed from Budapest to Amsterdam and it kept 'This Weekend' filter. Is it different on iOS or did you mean something else?

And if I think about our current only use case for this, when people select a date in the calendar app and open the discovery for that date, then when they select a different city, they may still want to start from the same day. They may only need to change the city because on that date they will be somewhere else.

So for me it would actually make sense to keep all filters and not reset them when new city is selected.

lemonboston:
A little correction: our start date is not really a filter at the moment, just the starting point to show in the event list.

lemonboston:
Fixed as part of updating an earlier pull request, so I am closing this.

PR #297, opened 2017-05-30
Fix loosing start date after city selection. #296

lemonboston:
Can be merged only after #287.


lemonboston:
Set story points to 2 as it took a little time to track it down.

lemonboston:
This has been fixed in the pull request #286 when updating that one.
So I am closing this pull request and the corresponding bug.

Issue #307, opened 2017-06-09
Change LoadResult type to throw Exception instead of the isSuccess() check


lemonboston:
@dmfs As you mentioned not being able to load something should be an Exception, so it's better to change this type.

PR #308, opened 2017-06-09
Change LoadResult to throw Exception. #307

lemonboston:


dmfs (approved):
Approving this. The actual review will be done with PR #286

Issue #281, opened 2017-05-23
Use button-up transition when opening event details


dmfs:
This also requires a custom fade transition to make the back transition to work as expected.

PR #311, opened 2017-06-20
Use bottom-up transition for showing details screen. #281

lemonboston:
Note: submitted this now for merging to 278, but we should go ahead with that instead and rebase this one on categories/master after that.

I used bottom-up for entering and fade-out for exiting. I tried with the bottom-up exit animation as well but that didn't look good. Tried to tweak it a bit as well, but couldn't make it nice quickly so I thought this fade-out looks good enough for now. I think if these animations represent an _okay_ version now then we should go ahead and just put an improvement ticket in the backlog if needed, which we can get back to after categories are implemented.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.2'
+        classpath 'com.android.tools.build:gradle:2.3.3'
 
         // NOTE: Do not place your application dependencies here; they belong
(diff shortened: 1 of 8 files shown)


lemonboston:
Changed the PR's base branch to `categories/master` and rebased onto that.

dmfs (approved):
Looks ok. Although the bottom-up animation is a but to fast IMO. We'll discuss this on slack with the team.
Also the feedback should be opened in another bottom-up animation, because it's on the navigation level. Please open an mx story.

Issue #312, opened 2017-06-21
Gitches for splash screen circle loader


lemonboston:
It starts a bit abruptly and with a short freeze and it is still shown in a little bit moved up position after the screen is gone and list is starting to show.

(Note: I think I have an easy fix for both.)

PR #313, opened 2017-06-21
Fix glitched of the circle progressbar on splash loading screen. #312

lemonboston:
This is based on #278, so that one should be merged to `categories/master` first, then this can be rebased.


lemonboston:
Changed base branch to `categories/master` and rebased onto that.

dmfs (approved):
Look ok.

Issue #315, opened 2017-06-22
(Categories branch) Duplicate location header on top of each other


lemonboston:
See: https://github.com/schedjoules/android-event-discovery-sdk/pull/286#issuecomment-310344510

One way to reproduce:
1. Open discovery list
2. Open an event details, rotate, rotate back, go back to list
3. Open location selection and select a different city
Result: both locations displayed on top of each other (see screenshot above)

wilfredstegeman:
I don't have this issue.

lemonboston:
It's only in a development branch, caused by recent changes, so this is not in any release and it will be fixed before it would get there. Marten created the labels `bug` and `live-bug` to distinguish between not-released and released bugs, I try to use them appropriately.

PR #316, opened 2017-06-22
Fix initialization check in EventListFragment to avoid duplicate Head…

lemonboston:
…erFragment. #315 

Note: this branch is based on #278 

@dmfs This is a one line change, after you've checked, I would like to propose again to move this to a base Fragment. We need to check 2 things in order to correctly determine if a Fragment is initializing, because `savedInstanceState` can be `null` when popping existing instance from backstack.
So in order to avoid having to think it through every time, and also avoid code duplication, we could move it to a base Fragment.
I see it as a framework functionality, so something which the framework could also provide us (maybe it even _should_ in this case). And such things for me are fine going into base classes.

So the proposal:
```java
class BaseFragment
{
    private boolean mIsNewInstance = true;

    public final View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
            @Nullable Bundle savedInstanceState) {
        View view = onCreateView(inflater, container, savedInstanceState, savedInstanceState == null && mIsNewInstance)
        mIsNewInstance = false;
        return view;
    }

    public abstract View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
            @Nullable Bundle savedInstanceState, boolean isInitializing);
}

```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.2'
+        classpath 'com.android.tools.build:gradle:2.3.3'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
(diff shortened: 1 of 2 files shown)


lemonboston:
Just to note the 2 scenarios how this bug happened:

1. When just going to details, and coming back. If `savedInstanceState == null` check is used, another HeaderFragement is added.  (this is the one I fixed earlier)
2. When rotating screen. If `mIsNewInstance` check is used, a new HeaderFragment is added besides the one that is recreated automatically by FragmentManager.

So this is why we need both checks.

lemonboston:
Changed PR's base branch to `categories/master` and rebased onto that.

dmfs (approved):

Issue #298, opened 2017-06-04
Update libraries to latest (java-api-client, http-client-essentials, iterators, optional)


dmfs:
The same commit should also update to iterators 1.4, because there was a breaking API change in Iterators which http-client-essentials depends on.

lemonboston:
I suppose we need a new release from `java-api-client` as well, right?
And I see that the directly declared `org.dmfs:rfc3986-uri:0.7` and `org.dmfs:optional:0.1` use `iterators:1.3.1`, so do we need to update them as well, first?

lemonboston:
Just found these nice ways to explore the dependency hierarchy, show the conflicts:
http://mrhaki.blogspot.hu/2014/08/gradle-goodness-getting-more-dependency.html

lemonboston:
Depends on https://github.com/dmfs/optional/issues/10 to have new `optional` lib with `iterators` 1.4.

lemonboston:
I suppose it makes sense to do this at the same time as updating to latest java-api-client, right? https://github.com/schedjoules/android-event-discovery-sdk/issues/300
So basically update everything at once (java-api-client, http-client-essentials, iterators, optional).
Does that sound right? I'd rename this task accordingly and remove #300.

I suggest we do this after 1st refactor-preparation phase of categories/master is merged to master and before starting to add categories loading.

dmfs:
yes that sounds right

lemonboston:
With updating
`http-client-essentials` to `0.12`
`java-api-client` to `0.8`
`optional` to `0.3`

`iterators` have conflicting version accross the dependencies:

`./gradlew eventdiscovery-sdk:dependencyInsight --configuration compile --dependency iterators`

```
org.dmfs:iterators:1.5 (conflict resolution)
+--- com.github.schedjoules.java-api-client:actions:0.8
|    \--- com.github.schedjoules:java-api-client:0.8
|         \--- compile
+--- com.github.schedjoules.java-api-client:eventsdiscovery:0.8
|    +--- com.github.schedjoules:java-api-client:0.8 (*)
|    +--- com.github.schedjoules.java-api-client:actions:0.8 (*)
|    \--- com.github.schedjoules.java-api-client:insights:0.8
|         \--- com.github.schedjoules:java-api-client:0.8 (*)
+--- com.github.schedjoules.java-api-client:insights:0.8 (*)
\--- com.github.schedjoules.java-api-client:schedjoules-api:0.8
[...]

PR #325, opened 2017-06-30
Update http-client-essentials, optional, java-api-client to latest. A…

lemonboston:
…djust code accordingly, new Box-related classes. Fix build. #298 

---

Branch is based on `categories/master`, could be merged to that before merging that to `master` or to `master` after that.

---

This pull request contains:

- Update libraries:
  - http-client-essentials to 0.12
  - uri-toolkit to 0.8
  - java-api-client to 0.9
  - optional to 0.3
- Update code accordingly, and
  - add several new Box-related classes, improve parcelling code in EventResultPageBox
- Fix build:
  - Use AppCompat versions of TextView, EditText for custom views (Lint error)
  - Fix a failing unit test (ListSectionChangeTest)
  - (edit:) Exclude support libs from dependencies to fix build (Lint error)


lemonboston:
After rebasing on `master`, with the new MF version the build failed because of more than one support lib version (25.0.1 and 25.3.x), so I had to exclude it and I've also fixed it for flexible-adapter.
Squashed and rebased.

eventdiscovery-sdk/src/sharedTest/java/com/schedjoules/eventdiscovery/testutils/EmptyResultPage.java
```diff
@@ -22,45 +22,35 @@
...
-        return true;
-    }
-
 
     @Override
-    public boolean isLastPage()
+    public Optional<ApiQuery<ResultPage<Envelope<Event>>>> previousPageQuery() throws IllegalStateException
     {
-        return true;
+        throw new IllegalStateException("No previous query");
```
dmfs: This should return an absent value.

eventdiscovery-sdk/src/sharedTest/java/com/schedjoules/eventdiscovery/testutils/EmptyResultPage.java
```diff
@@ -22,45 +22,35 @@
...
     }
 
 
     @Override
-    public ApiQuery<ResultPage<Envelope<Event>>> previousPageQuery() throws IllegalStateException
+    public Optional<ApiQuery<ResultPage<Envelope<Event>>>> nextPageQuery() throws IllegalStateException
     {
-        throw new IllegalStateException("No previous query");
+        throw new IllegalStateException("No next query");
```
dmfs: same here

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/serialization/boxes/ApiQueryBox.java
```diff
@@ -0,0 +1,95 @@
...
+        @Override
+        public ApiQueryBox[] newArray(int size)
+        {
+            return new ApiQueryBox[size];
+        }
+    };
+
+
+    // Note: method is needed instead of constant because of the generic parameter
+    public static <T> BoxFactory<ApiQuery<T>> factory()
```
dmfs: No, this method is not needed. Just make it an inner static class instead and call it like this
```java
new ApiQueryBox.Factory<>()
```

instead of this
```
ApiQueryBox.factory()
```
The result is the same, but we don't have this static method.

dmfs (changes requested):

lemonboston:
Made the corrections as per the review comments above, ready for check again.

dmfs (approved):

PR #331, opened 2017-07-04
Category labels fetching and storing in cache on splash loader screen…

lemonboston:
…. #326 

---

Additional notes:
- support lib version has been increased to 25.1.0 because arch libs use a class from that version
- as it turned out it's not needed after all to specify the `module` in the gradle `exclude` clauses, `group` is enough, so I removed them. Also removed support lib dependency declarations that didn't seem necessary.
- edit: this is outdated:
At first I created the the `ViewModel` with a public field, without constructor since I was unsure about that `Factory` that can be provided to `ViewModelProviders`. I used a new `Sealed` type to hold the value (it allows setting once and reading only after set).
But after I looked into the `ViewModelProviders` code, that factory usage seems okay, so I refactored it.
I've kept this last refactor commit separate, I think it may be easier to check that way what I was trying to solve. It's still not totally nice, for example I've added a static method. Also the `CategoriesCache(FragmentActivity activity)` constructor creates a double-wrapped `CategoriesCache`.
- accessing event categories will be done like this (tested, works well):
```
Iterator<Category> = new EventCategories(event, new CategoriesCache(getActivity()));
```


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/iterators/Present.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.Iterator;
+
+
+/**
+ * Decorator for {@link Iterator} with {@link Optional} elements to get only the present values.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class Present<E> extends AbstractBaseIterator<E>
```
dmfs, line 34: This should be called `Presents` or even better `PresentValues` because it's iterates the **values** of all present `Optional`s.
Also this would fit nicely into the optionals lib. Please create a PR and also add a `PresentValues`  `Iterable`

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/JavaUri.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.net.URI;
+
+
+/**
+ * {@link Uri} adapter for {@link URI}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class JavaUri implements Uri
```
dmfs, line 38: Just a comment:
I was considering to add an adapter like this to uri-toolkit. But I decided against it, because Java's URI implements RFC 2369 whereas uri-toolkit implements RFC 3986 so simply delegating to the `URI` methods wouldn't work.

Your solution handles this better than my idea. If you can replace the eager call to `javaNetUri.toString()` with a lazy adapter like `new URICharSequence(javaNetUri)`, this would be a candidate for `uri-toolkit` indeed.

lemonboston:
I changed `CategoriesCache` a bit, it looks better now, no static method, still not perfect but I think it's good enough.
( `new CategoriesCache(getActivitity()).cache(getActivity())` could be written, but it doesn't make sense.)

Squash-rebased, so there is only 1 commit now. Ready for review.

lemonboston:
As it turned the arch libs causes crash on older Android version with support lib 25.1.0.
The reason is not clear, but if we bounce it to 25.2.0 it runs.
Moving this back to in progress to look into it and update.

lemonboston:
This is now depends on #337 
and on https://github.com/schedjoules/java-api-client/pull/53 and getting a new tag from java-api-client to use instead of the hash.
After those it can be rebased on master.
Otherwise, ready to review.

lemonboston:
Update the api-client version and rebased on `categories/master`, changed base branch to that, so this can be reviewed and merged independently from #337.

dmfs:
You didn't address my request to rename the `Present` `Iterator` decorator yet.

lemonboston:
I must have missed your comment on this `Present` `Iterator` totally and I can't find it now.
What should I rename it to, what would be a better name for it?

lemonboston:
A small note: just noticed in another project that the error handling in `ContextActivity` is not correct, it allows to return `null`, I will get back to correct that later either on this branch or in a patch ticket.

lemonboston:
Missed the PR with the previous note, that is related to the next PR: https://github.com/schedjoules/android-event-discovery-sdk/pull/338

dmfs (approved):

Issue #333, opened 2017-07-06
Add note to code about possible FlexibleAdapter fix for the workarounds


lemonboston:
FlexibleAdapter author informed us 
(https://github.com/davideas/FlexibleAdapter/issues/277#issuecomment-313123331)
that there might be fixes for the issues we had, so our workaround could be removed potentially.
It's not worth spending time on it now because current version works, but we can add a comment to that particular code.

PR #334, opened 2017-07-06
Add note about potential FlexibleAdapter improvements. #333

lemonboston:


dmfs:
What's the point of this change? It's sufficient if we have a story to remind us about this TODO.

In future, I'd like to establish the following procedure for workarounds and todos in our code:

1. create a ticket about the todo or about revisiting the workaround once the original bug was fixed
2. refer to the ticket in a comment before committing the code

lemonboston:
Makes sense, thanks, I was unsure about how to handle, refer to this, is it enough to keep a TODO or create a ticket.
I've created the ticket: https://github.com/schedjoules/android-event-discovery-sdk/issues/335
And removed the long comment from the code in this PR, just pointing to the ticket instead. Ready for checking again.

dmfs (approved):

Issue #336, opened 2017-07-07
Crash on Android 4.3-4.4 when opening details screen


lemonboston:
```
07-07 12:52:22.687 E/AndroidRuntime: FATAL EXCEPTION: main
                                     Process: com.schedjoules.eventdiscovery.demo, PID: 3458
                                     java.lang.RuntimeException: Unknown interpolator name: pathInterpolator
                                         at android.view.animation.AnimationUtils.createInterpolatorFromXml(AnimationUtils.java:328)
                                         at android.view.animation.AnimationUtils.loadInterpolator(AnimationUtils.java:271)
                                         at android.view.animation.Animation.setInterpolator(Animation.java:391)
                                         at android.view.animation.Animation.<init>(Animation.java:255)
                                         at android.view.animation.AlphaAnimation.<init>(AlphaAnimation.java:40)
                                         at android.view.animation.AnimationUtils.createAnimationFromXml(AnimationUtils.java:116)
                                         at android.view.animation.AnimationUtils.createAnimationFromXml(AnimationUtils.java:114)
                                         at android.view.animation.AnimationUtils.createAnimationFromXml(AnimationUtils.java:91)
                                         at android.view.animation.AnimationUtils.loadAnimation(AnimationUtils.java:72)
                                         at android.support.v4.app.FragmentManagerImpl.loadAnimation(FragmentManager.java:892)
                                         at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1192)
                                         at android.support.v4.app.FragmentManagerImpl.removeFragment(FragmentManager.java:1392)
                                         at android.support.v4.app.BackStackRecord.run(BackStackRecord.java:758)
                                         at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1682)
                                         at android.support.v4.app.FragmentManagerImpl.executePendingTransactions(FragmentManager.java:594)
                                         at org.dmfs.android.microfragments.HostFragment.executeTransition(HostFragment.java:209)
                                         at org.dmfs.android.microfragments.HostFragment.onPigeonReturn(HostFragment.java:186)
                                         at org.dmfs.android.microfragments.HostFragment.onPigeonReturn(HostFragment.java:50)
                                         at org.dmfs.pigeonpost.localbroadcast.ParcelableDovecote$DovecotReceiver$1.run(ParcelableDovecote.java:103)
                                         at android.os.Handler.handleCallback(Handler.java:733)
[...]

Issue #320, opened 2017-06-28
Use Activity-like closing transition for details screen (-> update Fragment backgrounds)


lemonboston:
Because our Fragments don't have background, the standard Activity closing animation wouldn't look good currently. Update the backgrounds and the animation.

Note: `android:windowBackground` attribute might be good for this purpose in the Fragment layouts.

lemonboston:
Some part of the Activity-like opening transition has been removed as well in order to fix https://github.com/schedjoules/android-event-discovery-sdk/issues/336, so as part of this task we can put back them as well if needed, using older API level compliant versions.

PR #337, opened 2017-07-07
Fix crash on earlier Android caused by API 21 interpolator. #336

lemonboston:
@dmfs Please review.


dmfs:
Can't you just use another interpolator on these older platform versions instead of removing the entire code? Something like [decelerateInterpolator](https://developer.android.com/reference/android/view/animation/DecelerateInterpolator.html) could do the trick.

lemonboston:
Most of the removed code was not used actually, the only effective change is in `BottomUp` where the exit animation is changed from `schedjoules_activity_open_exit` to `microfragments_fade_exit`.
I just checked now and I see only a slight difference between the two.
So in order to keep focusing on delivering categories, I think we could get back to this in https://github.com/schedjoules/android-event-discovery-sdk/issues/320, I've already added a note there. Since the activity-like transition code was in master, we can easily restore them from history if needed, and then work out the 4.x compatible versions.

The reason why I ended up removing all the not used resources as well was that I got strange crashes even when they were not supposed to be used, it was not clear if it was caused by that or just some build mixup during development, but it seemed safer to just remove everything with `pathInterpolator` for now.

dmfs (approved):

PR #338, opened 2017-07-07
Add category labels to details screen. #332

lemonboston:
This based on #331 but otherwise ready to review.

After resolving all the older Android version related issues, I used this branch to smoke test on these versions, since this branch is on top of every other pending ones:
API Level 24 7.0 Nougat
API Level 23 6.0 Marshmallow
API Level 22 5.1 Lollipop
API Level 19 4.4 Kitkat
API Level 15 4.0 Ice Cream Sandwich

Found one more crash because of an atrribute color used in drawable, that was introduced in the branch, so fixed that as well.


lemonboston:
Rebased on `categories/ad-hocs/326-fetch-and-cache`.

dmfs (approved):

Issue #330, opened 2017-07-04
Test low memory (Activity killed, process death) scenarios


lemonboston:
In order to ensure correct working of the SDK, or at the minimum not crashing, in the rare scenarios when the system kills part or the whole app to save resources, as a developer I want a reliable way to test these.
---

The 2 methods usually suggested to test this - "Don't keep Activities", and kill the process in Android Studio - don't seem to work as expected for some reason, so we need to look into this more.

lemonboston:
TODO as part of this, check how categories cache is accessed after restoration.

lemonboston:
@dmfs I haven't been able to test it so far, if you have any ideas, suggestions please let me know.

These are the things I tried:

- Use "Don't keep Activities" phone debug option -> `MicroFragmentHostActivity` is not destroyed when I go to `LocationPickerActivity` or even outside the app with the Share button
- Kill the app from Android Studio with "Terminate Application" button while the app is in background, and then relaunch the app from recent apps -> the app starts with `MainActivity` instead of trying to restore `MicroFragmentHostActivity` with its `Fragment`s as multiple blog posts say it should be doing
- Use following adb commands to trigger low memory simulation: -> nothing happens, app stays intact
```
adb shell am send-trim-memory com.schedjoules.eventdiscovery.demo MODERATE
adb shell am send-trim-memory com.schedjoules.eventdiscovery.demo RUNNING_CRITICAL
```

-  I thought maybe the system is so smart that it knows about the retained Fragment and that alters the behvaiour, so I change it to non-retained -> nothing changed, the same behavior as above

It would be important to test this at this point because there is a good chance of a crash now when restoring the Activity and trying to reach the presumably emtpy `CategoriesCache`.
Not sure if it makes sense to try to fix that 'blindly' without actually seeing a crash.

lemonboston:
- also tried to launch `LocationPickerActivity` not as `startActivityForResult` but with `startActivity` so that maybe that's what kept the `MicroFragmentHostActivity` live, but this didn't change anything either

lemonboston:
This was all done on Nexus 5 device, but now I tried on API 22 Emulator, the same behavior.

lemonboston:
I haven't been able to test it so far, so in order to move forward, I've applied a 'blind' fix for this by saving the cache to `savedInstanceState` in #340. 
I keep this ticket open but move it back to Backlog for now.

lemonboston:
[...]

PR #340, opened 2017-07-11
Save and restore Categories cache when Activity is recreated after sy…

lemonboston:
…stem killed it. #339 

---

I've added saving and restoring `CategoriesCache` from `savedInstanceState` in `MicroFragmentHostActivity`. This is to handle the case when Activity would be restored after it had been killed, it wouldn't be needed for just configuration changes. I've just used the `savedInstanceState != null` check nevertheless, to not complicate the logic with extra flags. Since serialization doesn't really happen on config change, I think it's fine.

I haven't been able to test this real scenario though, https://github.com/schedjoules/android-event-discovery-sdk/issues/330 will be kept open for that. So this PR is for applying a 'blind' fix for that possible crash, so that we can move forward.

I did test the parcelling added in this PR by forcing it.

### DIFF
--- a/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/activities/MicroFragmentHostActivity.java
+++ b/eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/activities/MicroFragmentHostActivity.java
@@ -21,9 +21,12 @@ import android.os.Bundle;
 import android.support.annotation.NonNull;
 
 import com.schedjoules.eventdiscovery.framework.common.BaseActivity;
+import com.schedjoules.eventdiscovery.framework.common.CategoriesCache;
 import com.schedjoules.eventdiscovery.framework.serialization.Keys;
+import com.schedjoules.eventdiscovery.framework.serialization.boxes.CategoriesBox;
 import com.schedjoules.eventdiscovery.framework.serialization.boxes.ParcelableBox;
 import com.schedjoules.eventdiscovery.framework.serialization.commons.Argument;
+import com.schedjoules.eventdiscovery.framework.serialization.commons.FluentBundle;
 import com.schedjoules.eventdiscovery.framework.serialization.commons.IntentBuilder;
 
 import org.dmfs.android.microfragments.MicroFragment;
@@ -49,6 +52,13 @@ public final class MicroFragmentHostActivity extends BaseActivity
     protected void onCreate(Bundle savedInstanceState)
     {
         super.onCreate(savedInstanceState);
+
+        if (savedInstanceState != null)
+        {
+            // Categories cache restoration is needed for the case when Activity is restored after being killed
+            // (It wouldn't be needed for config change - that's handled by ViewModel)
+            new CategoriesCache(new Argument<>(Keys.CATEGORIES, savedInstanceState).get()).cache(this);
+        }
 
         // the BackDovecote receives Pigeons with the result of the BackTransition.
         mBackDovecote = new BooleanDovecote(this, "backresult", new Dovecote.OnPigeonReturnCallback<Boolean>()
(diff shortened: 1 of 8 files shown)


lemonboston:
@dmfs One more note to this: I know that the fact that we still need to save the cache to savedState even though we use `ViewModeel` kind of takes away the supposed simplicity of it. So I don't consider this an ideal solution, I mean even using `ViewModel`. I like you're ideas to handle states uniformly and go towards immutable views much better, too, so in the long run we can figure out architectures for that (I have some ideas already..).

But at this point this `ViewModel` for the categories cache looks good enough for me, to have a first version of the feature.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/activities/MicroFragmentHostActivity.java
```diff
@@ -87,9 +97,9 @@ protected void onSaveInstanceState(Bundle outState)
     {
         super.onSaveInstanceState(outState);
 
-        // TODO Possibly create mutable State abstraction that holds all state for the Activity.
-        // TODO Key should not be used like this:
-        outState.putParcelable(Keys.MICRO_FRAGMENT_HOST.name(), new ParcelableBox<>(mMicroFragmentHost));
+        new MutableBundle(outState)
```
dmfs: Not a fan of this. IMO it's better/cleaner/simpler to use a `BundleBuilder` and add the result "manually" (as a nested bundle) to `outState` using `putParcelable`.
This `MutableBundle` thing doesn't really add anything and the name "MutableBundle" already indicates its pointlessness, because a `Bundle` is already mutable.

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/activities/MicroFragmentHostActivity.java
lemonboston: So that could be something like this:
```java
        outState.putParcelable("schedjoules.state",
                new BundleBuilder()
                        .with(Keys.MICRO_FRAGMENT_HOST, new ParcelableBox<>(mMicroFragmentHost))
                        .with(Keys.CATEGORIES, new CategoriesBox(new CategoriesCache(this)))
                        .build());
```

My problem with this is that we start to expose the details of our serialization mechanism, I would like to avoid calling `putParcelable` directly or probably calling anything on `Bundle`.
So there is that `"schedjoules.state"` string key here now, which is not a `Key`. We can make it one, but then we would call `.name()` on it, which is not how it is supposed to work.

This is how we read the value in `onCreate()` now:
```java
new Argument<>(Keys.CATEGORIES, savedInstanceState).get()
```
This is straightforward, in-line with how we use it for Activity Intent or Fragment args. No need to think. With the version above, I wouldn't be sure how to access the values.

So I think there is great value in not exposing how we put and retrieve values to and from Bundles, especially nesting, and just using `Key<T>` accordingly.

`MutableBundle` is not the best name, I agree. The other option I thought is putting back that `Parameter.into(Bundle)` interface..  not sure if that's better. That couldn't be chained.

How about `FluentBundle` for the class? We are basically wrapping a Bundle here for adapting to our serialization and also making it fluent, so it seems fitting.

lemonboston:
I needed to use this in the filter ui branch as well, so I've renamed `MutableBundle` to `FluentBundle` and the method to `put` instead of `add` and rebased.

I think that how to do this may be part of a bigger topic of state saving, I have some ideas about that to discuss but I'd postpone that for later.

lemonboston:
I've rebased this onto `master`, we need this one in the next release. (Just found a way to test this accidentally during the regression test... it causes a crash)

dmfs (approved):

Issue #341, opened 2017-07-11
Create category filter UI bar


lemonboston:
Create the filter drop-down widget with only the categories, so no date or settings, and categories are just hardcoded dummies.
Widget is interactive, so drops down on tap, highlights selected items, collapses on tap on the header again, has a CLEAR item for clearing.
When dropping down, it is either an overlay on the list or pushes it down, whichever looks easier.

---

(Personal note: I think pushing the list down, so that filter can be left open and every change in it triggers reload of the list below without the need to close the filter could be a nice, interactive UX. Given that expanded filter doesn't cover the whole screen.)

wilfredstegeman:
The most common filter is an overlay on the list when dropping down, which lso has my preference. The option of direct filtering of the results is indeed a nice interactive UX. Can you also achieve that with an overlay? And to have this effect also with more content in the dropdown we could maximize the size of the overlay, so you have to scroll through the results?

dmfs:
In general I think an overlay is the more intuitive and simpler solution. We can certainly implement interactive UX with instant updates using the overlay pattern as well.

That being said, I can imagine using a push-down solution in combination with the quick-return pattern (see https://www.youtube.com/watch?v=PL9s0IJ9oiI). So when you expand the filter, it will scroll out of the view when you scroll down but it will reappear instantly when you scroll up again.

In general I think the quick filter should list only the most common or most frequently used options. If you need more there should be an "advanced filtering" button which presents you with the full list.

lemonboston:
Right, I didn't realize that we would need quick-return as well for the push-down, otherwise we need to scroll to the top to see the filter. (A simpler vid about quick-return: https://www.youtube.com/watch?v=BwLjMMIWNQU)
This would probably complicate the solution so I will aim for the overlay first as you guys suggest. 

Updating the list on each category tap or only when filter is collapsed is probably not a big difference, I will see.

lemonboston:
@wilfredstegeman @dmfs Could you please provide NL, DE translations for "Category" for the filter title?

wilfredstegeman:
NL: Categorie
________________________________

dmfs:
DE: Kategorie

lemonboston:
Thanks, we also need to translate "CLEAR" which is the bottom item on the drop-down filter to clear and collapse the filter.
The mockup:
https://projects.invisionapp.com/d/main#/console/10528802/228235134/preview#project_console

Could you please provide NL, DE versions?

[...]

PR #343, opened 2017-07-18
Create categories filter UI (dummy). #341

lemonboston:
This pull request contains the first form of the categories filter UI widget.

Outstanding issues, questions that may need to be addressed in separate tickets before this can be considered ready for a public release: https://github.com/schedjoules/general-issues/issues/197

Note, I may come back to update (and squash) this branch while I will be working on wiring in the list update, so please check, update each time.


dmfs:
please rebase to master and resolve conflicts, I don't think we need categories/master anymore.

lemonboston:
I've reset `categories/master` to `master` and rebased this one on top.

dmfs:
For now I have one question: Instead of having these `CategoryOption` objects, wouldn't it be easier if you just have twp `Iterables` one for all available Categories and another one for the selected categories? The first one wouldn't have to change and the later one is probably easier to manage as well.

lemonboston:
Two `Iterable`s would certainly be an option, but I like that `CategoryOption` fits well as the model for `FilterItemView`, so it get's updated with a category and whether that is selected or not.
And similarly `Iterable<CategoryOption>` fits well as the 'model' for the whole filter (there is private `update` method with that. With the 2 `Iterable`s we need to keep them in sync, if let's say we will handle language change.

eventdiscovery-sdk/src/main/res/layout/schedjoules_view_filter_item_clear.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout>
+
+    <FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
dmfs: What's this FrameLayout for?

lemonboston:
Replaced margins with paddings as discussed.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/categoryoption/NegateSelected.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.schedjoules.client.eventsdiscovery.Category;
+
+
+/**
+ * {@link CategoryOption} decorator negating {@link CategoryOption#isSelected()}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class NegateSelected implements CategoryOption
```
dmfs: How about the name `SelectionComplement`?

eventdiscovery-sdk/src/main/res/layout/schedjoules_view_divider.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
dmfs: What's the FrameLayout good for?

dmfs (changes requested):
I do have a couple of concerns regarding the design. For the sake of progress I'll accept them since it seems to work nevertheless.

eventdiscovery-sdk/src/main/res/layout/schedjoules_view_filter_item_clear.xml
lemonboston: Nothing, copied from the item layout and left in mistakenly. I removed it.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/categoryoption/NegateSelected.java
lemonboston: Renamed to `SelectionToggled` as suggested on chat.

eventdiscovery-sdk/src/main/res/layout/schedjoules_view_divider.xml
lemonboston: I remembered as if I had some problem with simply using a View as root, but now I tried and it works, so I changed it.

lemonboston:
Ready for check again.
The design in `EventFilterView` could certainly be improved later, imo, too.

dmfs (approved):

Issue #344, opened 2017-07-19
Crash when opening location picker


dmfs:
> I get the following crash when selecting the ALL LOCATIONS item on top. So do several users.
With 1.9 I could not repoduce this. Sometime it doesn't crash though - no idea yet on why.
It shows the ENTER LOCATION/MY LOCATION for about 500ms before crashing.
Any ideas?
```
java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 927544 bytes
	at android.app.ActivityThread$StopInfo.run(ActivityThread.java:3776)
	at android.os.Handler.handleCallback(Handler.java:751)
	at android.os.Handler.dispatchMessage(Handler.java:95)
	at android.os.Looper.loop(Looper.java:154)
	at android.app.ActivityThread.main(ActivityThread.java:6123)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:867)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:757)
Caused by: android.os.TransactionTooLargeException: data parcel size 927544 bytes
	at android.os.BinderProxy.transactNative(Native Method)
	at android.os.BinderProxy.transact(Binder.java:615)
	at android.app.ActivityManagerProxy.activityStopped(ActivityManagerNative.java:3700)
	at android.app.ActivityThread$StopInfo.run(ActivityThread.java:3768)
	... 7 more
```

dmfs:
I think the problem is a large amount of data in the saved state of the activity.
The quick fix is probably to not store the results in the saved state and let it reload the list on configuration changes.

dmfs:
After looking into the code it seems it's only the initial page which is retained. So another solution could be to let the EventListFragment take care of loading the initial events or to reduce the number of events on the initial result page.

lemonboston:
I am checking this. We don't actually save the first `ResultPage` explicitly but since it is a `Fragment` argument the framework saves it to be able to restore the `Fragment`.

It's strange that it happens when opening the location picker, I assume that it means that the device is low on memory and Activities get killed instantly when they are left. The same as the "Don't keep Activities" developer setting.

[...]

PR #345, opened 2017-07-21
Ignore Event descriptions when used in discovery lists, but cache it.…

lemonboston:
… Load from cache or network when opening details screen. Remove first loader MF. #344


eventdiscovery-sdk/src/main/AndroidManifest.xml
```diff
@@ -74,6 +74,16 @@
         </service>
 
         <service
+                android:name=".framework.services.BasicEventService"
+                android:exported="false"
+                android:stopWithTask="true"
+                android:enabled="true">
+            <intent-filter>
```
dmfs: This service should not be bound by action, so no need for an intent filter.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/utils/cache/ExpiringLruCache.java
```diff
@@ -0,0 +1,219 @@
+/*
+ * Copyright 2017 SchedJoules
```
dmfs: Check copyright and license.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/services/EventService.java
```diff
@@ -0,0 +1,107 @@
...
+    Optional<Event> cachedEvent(String uid);
+
+    final class FutureConnection implements FutureServiceConnection<EventService>
+    {
+        private final FutureLocalServiceConnection<EventService> mDelegate;
+
+
+        public FutureConnection(Context context)
+        {
+            mDelegate = new FutureLocalServiceConnection<>(context, new Intent("com.schedjoules.EVENT").setPackage(context.getPackageName()));
```
dmfs: Bind this service by class instead of action. The service is not supposed to be replaced or overridden by partners.

dmfs (changes requested):

lemonboston:
Changed `EventService` to use explicit Intent and removed the copyright from `ExpiringLruCache`.

dmfs (approved):

PR #348, opened 2017-07-21
Use real Categories for the filter UI. #347

lemonboston:
Notes for after you have looked at the code:

Question: does it look good to get the main categories by counting the number of path elements? (In `FilterCategories`). 
Strangely the first path element is a "" (empty string) that's why it's 3 for main and 4 for subcategory. Is that a bug in `uri-toolkit`?

About `EagerCategories` - it's needed because those evaluations can take long, supposedly because of the `Uri` operations (even though it's just a String underneath, when we calll `new Text(uri).toString()` it parses and then concatenates again).
It decreased the time of populating the filter from ~100ms to ~20ms.
This is also why I put the `filteredCategories()` on `Categories` for the sake of simplicity now instead of using it separately. (Could be put into a separate 'cache', i.e. `ViewModel`).

In general about eager preparation or 'pre-caching' in constructors, I think it's fine. Yegor writes in the book that lazy composition is good because we have control over when to evaluate. Which is true. He also mentions the possible `Cached` decorator. What I think he doesn't face if he doesn't work on UI that we also need to 'pre-cache' stuff that are costly to run later on UI thread.
So I would add this to the 'pattern' in general:

```
interface Foo
{
    Something get()
}

class Cached implements Foo
{
   // the usual caching decoration
}

class PreCached implements Foo
{
   PreCached(Foo delegate)
   {
        mDelegate = new Cached(delegate);
        mDelegate.get();
   }
}
```
Calling safe methods on the delegate in the constructor for this preparation purpose looks okay to me.


dmfs:
Just a quick comment on the eager evaluation: I actually agree that constructors should not contain code other than assignments and object instantiations. If you need to "process" something in order to create an object you want a "factory" to build the object. That way it's perfectly clear that the object needs some computation before it can be created.

Regarding `Uri` operations, I'm already addressing this issue with the redesign. I hope to get rid of some of the overhead while still maintaining a good design.

lemonboston:
Added a TODO and a tracking ticket about changing that Eager to a factory:
https://github.com/schedjoules/android-event-discovery-sdk/issues/349

lemonboston:
Rebased on `categories/341-filter-ui`

dmfs (approved):
just squashing

PR #354, opened 2017-07-25
Add category labels to details screen. #353

lemonboston:
Since I had to resolve meaningful conflicts during rebasing categories/master to master and make additional changes as well, I prepared a separate branch for this (categories/master is untouched).

This now contains all the changes to add categories to details screen in a single squashed commit, although the diff of what I just added and changed cannot be seen separately.
So I list them here:
`FirstResultPageHolder`
`EventListLoaderMicroFragment` - first result page is now added to this holder and not to the outgoing bundle
`EventListFragment` line 143 - if first result page is available in the holder on initialization, show that.

I know that `FirstResultPageHolder` is not nice in this way but I don't have a good idea atm how to design it to work well for the clearing, too. I thought that in this form it is at least compact and easy to understand,  and the workaround nature of it is documented.


lemonboston:
When this has been merged to master I can rebase the categories branches onto master and we can remove categories/master.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/common/FirstResultPageHolder.java
```diff
@@ -0,0 +1,66 @@
...
+
+
+/**
+ * Holds the first result page loaded on splash screen in memory to be able to access it from the list screen.
+ * <p>
+ * Note: This is workaround that may live for a while but could potentially be removed later if caching is worked out.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class FirstResultPageHolder extends ViewModel
```
dmfs, line 40: I'll accept this as a temporary workaround, but in general I think we don't need this. Please create a story to replace it unless it is obsoleted by other changes.

dmfs (approved):

Issue #355, opened 2017-07-25
Reduce complexity of ApiService code


dmfs:
The ApiService code is quite verbose and mostly the same for all partners. It could be reduced to something like this:

```java
return new SdkApi(context, new StringAccessToken("0443a55244bb2b6224fd48e0416f0d9c"));
```
Information in `VersionedProduct` could be loaded automatically from the manifest.

lemonboston:
Just an obvious not-to-forget note that the wiki needs to updated, too, when this is closed.

PR #356, opened 2017-07-25
Simplify ApiServices by providing defaults which satifsy the needs of…

dmfs:
… most if not all partners. implements #355

Note: `DefaultApi` became an inner class of `AbstractApiService` because that appeared to be the only reasonable location, considering that everything under framework is considered to be SDK internal and may change at any time. `DefaultApi` needed a place where it's safe from being moved during a refactoring.

### DIFF
--- a/eventdiscovery-demo/src/main/java/com/schedjoules/eventdiscovery/demo/ApiService.java
+++ b/eventdiscovery-demo/src/main/java/com/schedjoules/eventdiscovery/demo/ApiService.java
@@ -6,20 +6,10 @@ import android.support.annotation.NonNull;
 
 import com.schedjoules.client.Api;
 import com.schedjoules.client.SchedJoulesApi;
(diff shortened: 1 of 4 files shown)


dmfs:
@lemonboston I was in a mood to implement this, but this PR is low prio.

lemonboston (approved):
Looks good to me.
Checked if packageName and application ID is always the same even if the manifest has different `package` attribute but they are, that attribute is only acts as a prefix for classes, the merged manifest contains the application id. So in short, it's good :)

Could be simplified even more with an extra constructor:
```
new DefaultApi(context, "0443a55244bb2b6224fd48e0416f0d9c", BuildConfig.LOG_REQUESTS)
```
The log tag is probably not important to customize.

I didn't really get why `DefaultApi` needs to be inner class, can't it be just in the same package as top level class?

dmfs:
I didn't want to add the log parameter because the request log should not really be used by partners.

Issue #352, opened 2017-07-25
SocketTimeoutException causes crash


lemonboston:
@lemonboston commented on [Mon Jul 24 2017](https://github.com/schedjoules/general-issues/issues/199)

I got this several times today while testing the new categories filtered discovery request, but from the stacktrace it has nothing to do with it.

@dmfs Can it be a bug in `http-client-essentials` that a timeout results in a `RuntimeException` propagated to the client?
(I haven't myself looked into it much yet)

```
07-24 13:55:04.758 E/AndroidRuntime: FATAL EXCEPTION: Thread-17470
                                     Process: com.schedjoules.eventdiscovery.demo, PID: 24605
                                     java.lang.RuntimeException: Can't get response code
                                         at org.dmfs.httpessentials.httpurlconnection.HttpUrlConnectionResponse.status(HttpUrlConnectionResponse.java:70)
                                         at com.schedjoules.client.eventsdiscovery.http.GetRequest.responseHandler(GetRequest.java:84)
                                         at org.dmfs.httpessentials.decoration.HeaderDecorated.responseHandler(HeaderDecorated.java:75)
                                         at org.dmfs.httpessentials.decoration.HeaderDecorated.responseHandler(HeaderDecorated.java:75)
                                         at com.schedjoules.client.http.GzipRequest.responseHandler(GzipRequest.java:116)
                                         at org.dmfs.httpessentials.decoration.HeaderDecorated.responseHandler(HeaderDecorated.java:75)
                                         at org.dmfs.httpessentials.decoration.HeaderDecorated.responseHandler(HeaderDecorated.java:75)
                                         at org.dmfs.httpessentials.decoration.HeaderDecorated.responseHandler(HeaderDecorated.java:75)
                                         at org.dmfs.httpessentials.httpurlconnection.PlainHttpUrlConnectionExecutor.execute(PlainHttpUrlConnectionExecutor.java:81)
                                         at org.dmfs.httpessentials.httpurlconnection.utils.executors.BottomBranded.execute(BottomBranded.java:56)
                                         at org.dmfs.httpessentials.httpurlconnection.utils.executors.BottomBranded.execute(BottomBranded.java:56)
                                         at org.dmfs.httpessentials.httpurlconnection.HttpUrlConnectionExecutor.execute(HttpUrlConnectionExecutor.java:68)
                                         at org.dmfs.httpessentials.executors.retrying.Retrying.retryingExecute(Retrying.java:68)
[...]

PR #358, opened 2017-07-26
Update java-api-client to 0.11 to fix crash for SocketTimeout-s. #352

lemonboston:


dmfs (changes requested):
Please also update http-client-essentials to 0.12.1

dmfs (approved):

PR #368, opened 2017-07-27
Reload list on category selection. #346

lemonboston:


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListFragment.java
```diff
@@ -94,6 +110,11 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
         mMenu = new EventListMenu(this);
         setHasOptionsMenu(true);
 
+        if (savedInstanceState != null)
+        {
+            mSelectedCategories = new Argument<>(Keys.OPT_FILTER_CATEGORIES, savedInstanceState).get();
```
dmfs: Why is `mSelectedCategories` an `Optional`?

Instead of holding it as an optional you could just do 
```
mSelectedCategories = new Argument<>(Keys.OPT_FILTER_CATEGORIES, savedInstanceState).get().value(EmptyIterable.instance());
```
to make it an empty `Iterable` if no selection was provided. "no selection" means "empty selection" after all, doesn't it? This would also make it slightly easier down the road.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
```diff
@@ -80,56 +90,66 @@ public EventFilterView(Context context, @Nullable AttributeSet attrs)
...
+        for (CategoryOption categoryOption : mCategoryOptions)
         {
-            @Override
-            public void onClick(View v)
+            mItemViews.get(i++).update(categoryOption);
+            if (!hasSelection)
             {
-                mPopup.dismiss();
-                update(new ClearedSelection(mCategoryOptions));
+                hasSelection = categoryOption.isSelected();
```
dmfs:
```
if (!hasSelection)
{
    hasSelection = categoryOption.isSelected();
}
```
can easily be replaced by 
```
hasSelection = hasSelection | categoryOption.isSelected();
```
Which also expresses the intent much clearer IMO.

dmfs (changes requested):

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/eventlist/EventListFragment.java
lemonboston: Yes, it is simpler with that indeed, updated.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
lemonboston: Updated

dmfs (approved):
Just squashing

Issue #369, opened 2017-07-27
UI becames sluggish with the real category filter


lemonboston:
Tapping on filters don't update them immediately but with some delay.
Also coming back from details screen has a delay as well when pressing back.

PR #370, opened 2017-07-27
Fix endless nesting of composed costly objects when selecting categor…

lemonboston:
…ies. #369 

---

The every time re-parsing of string uris are still costly but that wasn't the main problem. It would be nice though to have that solved as well.


dmfs (approved):

dmfs (approved):
Just squashing commits

Issue #380, opened 2017-07-28
'Category' filter title needs double tap to open sometimes


lemonboston:
Steps: open discovery, tap on 'Category' to expand, tap next to 'Category' on the title bar to dismiss the drop-down, then tap on 'Category' again. It doesn't react to first tap, only to the second.

PR #381, opened 2017-07-28
Remove popup dismiss on outside touch and extend title touch on the b…

lemonboston:
…ar, add ripple effect. #380


dmfs (approved):
Just squashing commits

PR #383, opened 2017-07-28
Show up arrow next to 'Category' when drop-down is expanded. #382

lemonboston:


dmfs (approved):
Just squashing commits.

PR #386, opened 2017-07-28
Change CLEAR button to have Borderless style. #385

lemonboston:


lemonboston:
The touch ripple doesn't extend to full height of the CLEAR item even though there is no padding for the button. I guess it would be nice to fix that.

dmfs (approved):
Just squashing commits.

PR #388, opened 2017-07-28
Change CLEAR button to not collapse the filter. #387

lemonboston:


lemonboston:
Probably there are some opportunities in `EventFilterView` now for code extraction to simplify it, like make a `SmartView` out of the category items or the whole dropdown. It didn't present itself clearly for me though now so I didn't want to spend time with it, but let me know if I should do it.
(There is some complexity now with the separation of initialization and update and with the different click listeners, what to do on each of them.)

dmfs (approved):

PR #399, opened 2017-08-03
Add SDK to the default user agent, implements #398

dmfs:


lemonboston (approved):
Looks good to me.

The debugged user-agent header, just to look it up easily later:
```
com.schedjoules.eventdiscovery/1.12-1-g54a323a com.schedjoules.eventdiscovery.demo/1.12-1-g54a323a schedjoules-api/0.11 httpurlconnection-executor/0.12.1 Dalvik/2.1.0 (Linux; U; Android 6.0.1; Nexus 5 Build/M4B30Z)
```

(The order of them can be corrected later in http-client-essentials as we discussed on chat, server doesn't depend on that.)

Issue #402, opened 2017-08-07
Change "clear" filter button to "show all"


dmfs:
"Show all" becomes another "category" which is mutually exclusive with all the other categories. It's selected if no other category is selected and clicking it will select it and unselect all categories.

"Show all" will be on top of all categories.

PR #408, opened 2017-08-09
replace "clear filter" with an "All" "category". Implements #402

dmfs:


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
```diff
@@ -112,6 +113,17 @@ private void init(Iterable<CategoryOption> categoryOptions)
         // Populate the drop-down
         mDropDown.removeAllViews();
         LayoutInflater inflater = LayoutInflater.from(getContext());
+
+        inflater.inflate(R.layout.schedjoules_view_divider, mDropDown, true);
+
+        // Add show all item
+        SchedjoulesViewFilterItemBinding clearItem = DataBindingUtil.inflate(inflater, R.layout.schedjoules_view_filter_item, mDropDown, true);
```
lemonboston: Variable name here is incorrect.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
```diff
@@ -112,6 +113,17 @@ private void init(Iterable<CategoryOption> categoryOptions)
...
+        clearItem.schedjoulesFilterItemLabel.setText(R.string.schedjoules_filter_show_all);
+        clearItem.getRoot().setOnClickListener(new ShowAllSelectListener());
+        clearItem.getRoot().setFocusable(true);
+        mShowAllItemView = new Highlightable(clearItem.schedjoulesFilterItemLabel);
+        mShowAllItemView.update(true);
```
lemonboston: I suggest moving `mShowAllItemView.update()` from here and from line 218 to `updateItems()` method.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
```diff
@@ -77,6 +77,7 @@
     private CategorySelectionChangeListener mCategorySelectionChangeListener;
     private LinearLayout mDropDown;
     private List<FilterItemView> mItemViews;
+    private SmartView<Boolean> mShowAllItemView;
```
lemonboston: I think it would be worth creating a `SmartView<FilterState>` for this item, so that we only need to update it with `mFilterState`, a little bit less thinking and complexity in this `EventFilterView`, no need to use `true` or `false`.
The class of course would delegate to `Highlightable` and use the negated `hasSelection()`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
```diff
@@ -112,6 +113,17 @@ private void init(Iterable<CategoryOption> categoryOptions)
...
         mDropDown.removeAllViews();
         LayoutInflater inflater = LayoutInflater.from(getContext());
+
+        inflater.inflate(R.layout.schedjoules_view_divider, mDropDown, true);
+
+        // Add show all item
+        SchedjoulesViewFilterItemBinding clearItem = DataBindingUtil.inflate(inflater, R.layout.schedjoules_view_filter_item, mDropDown, true);
+        clearItem.schedjoulesFilterItemLabel.setText(R.string.schedjoules_filter_show_all);
+        clearItem.getRoot().setOnClickListener(new ShowAllSelectListener());
+        clearItem.getRoot().setFocusable(true);
```
lemonboston: Could focusable be applied to the whole `EventFilterView` instead of each item? Could that work as well with the outside touch handling?

dmfs: You need to make the items focusable which are supposed to be focusable. Making each item focusable allows to navigate from one item to the next one with up and down arrow key.
Not sure what you mean with the outside touch handling in this context.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
dmfs: fixed

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
dmfs: I did this, although I'm not really sure if that makes it easier.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
dmfs: done

dmfs:
@lemonboston rebased on master and ready for review again

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/FilterStateItemView.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.schedjoules.eventdiscovery.databinding.SchedjoulesViewFilterItemBinding;
+import com.schedjoules.eventdiscovery.framework.filter.categoryoption.CategoryOption;
+import com.schedjoules.eventdiscovery.framework.filter.filterstate.FilterState;
+import com.schedjoules.eventdiscovery.framework.utils.smartview.SmartView;
+import com.schedjoules.eventdiscovery.framework.widgets.Highlightable;
+
+
+/**
+ * Represents the View for a category filter state item.
```
lemonboston: I think this is not a general view for an item view, it is specifically for the "All" item because of the inverse logic of highlighting in it, that's why I thought in the first place that it can be worth creating, to contain that. Thinking about inverse logic, `!` and `true` and `false` in an already complex class can be error prone imo. So that was the motivation.
So this could be something like `AllSelectionItemView`.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/FilterStateItemView.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author Marten Gajda
+ */
+public final class FilterStateItemView implements SmartView<FilterState>
+{
+    private final SchedjoulesViewFilterItemBinding mBinding;
+    private final SmartView<Boolean> mHighlightableLabel;
+
+    private CategoryOption mCategoryOption;
```
lemonboston: Not used.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/FilterStateItemView.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+    public FilterStateItemView(SchedjoulesViewFilterItemBinding binding)
+    {
+        mBinding = binding;
```
lemonboston: No need for the field.

lemonboston (changes requested):
Would you please squash and rebase after the fixes?

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
lemonboston: I thought that focusable had something to do with making the outside touch work as intended, nevermind then.

dmfs: That's why the popup itself is focusable. It ensures that background clicks are not propagated to the view underneath.

lemonboston: I see, thanks, so it did have something to do with the touches as well. I get that items has to be focusable, too, for navigation.

dmfs:
@lemonboston done, ready for re-review

lemonboston (approved):
Looks good to me

Issue #351, opened 2017-07-24
Replace Mapped(Optional) class of SDK with optional lib's Mapped(Optional)


lemonboston:
Use `org.dmfs.optional.decorators.Mapped` instead of `com.schedjoules.eventdiscovery.framework.utils.optionals.Mapped` and remove it.

dmfs:
optional 0.4 has just been released

dmfs:
Optional 0.5 is available now

lemonboston:
If latest `optional` uses newer `iterators` as well, then TODO replace `EmptyIterator` from that.

PR #409, opened 2017-08-09
Replace SDK `Mapped` with `optionals` `Mapped`. Implements #351

dmfs:
Also in this commit:
Replace deprecated `Iterator` decorators
Remove `First` in favour of `Next` from `optionals` for it does exactly the same and you can't really say if an element is the first element of an iterator. You can only say "give me the next one".


lemonboston (approved):
Looks good to me.

PR #410, opened 2017-08-09
Sort categories before showing them. Implements #400

dmfs:


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
```diff
@@ -104,8 +107,15 @@ protected void onFinishInflate()
 
         mCategorySelectListener = new CategorySelectListener();
 
-        init(new UnselectedCategories(
-                new ContextArgument<>(Keys.CATEGORIES, getContext()).get().filterCategories()));
+        init(new UnselectedCategories(new Sorted<>(
```
lemonboston, line 110: I think it's better if `Categories.filterCategories()` already return them sorted. So sorting can be put to `EagerCategories` constructor (which later may become a factory). (You don't necessarily need the `Sorted` decorator there since there is a `List` there which can be sorted, but can of course be kept as well since already implemented.)
`EagerCategories` is the sort of single 'entry point' of the categories so this would make sure wherever we access them, they are sorted, so it doesn't have to be a presentation logic concern and extra code in the already complex `EventFilterView`.

lemonboston (approved):
I don't want to hold up PRs with opinionated comments, so I checked this again and since it works fine and code looks correct to me, I mark it 'approved'.

Issue #412, opened 2017-08-14
"All" filter button always highlighted in dark theme


dmfs:

lemonboston:
I couldn't reproduce it with Nexus5-Android6 and Emulator-Android7.

PR #413, opened 2017-08-17
Ensure the filter views are invalidated explicitly to redraw them. Fi…

dmfs:
…xes #412


lemonboston (approved):
I couldn't reproduce the bug myself but the fix looks harmless in any way.

PR #414, opened 2017-08-17
Send insights from categories filter. Implements #404

dmfs:


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/insights/Opened.java
```diff
@@ -0,0 +1,61 @@
...
+ * Category Filter opened step.
+ *
+ * @author Marten Gajda
+ */
+public final class Opened extends AbstractStep
+{
+    private final FilterStep mDelegate;
+
+
+    public Opened(FilterStep delegate)
```
lemonboston: Maybe have default ctor which creates the `FilterStep` for this class and all others to have shorter usage in `EventFilterView`? I know these are decorators and with those ctors they wouldn't be but it would be fine imo.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
```diff
@@ -229,6 +236,11 @@ public void onClick(CategoryOption categoryOption)
             mTitleView.update(mFilterState);
             mShowAllItemView.update(mFilterState);
 
+            new InsightsTask(getContext()).execute(
+                    categoryOption.isSelected() ?
```
lemonboston: Maybe have something like `new Tapped(categoryOption)` to hide all this?

lemonboston (approved):
Code looks correct and the design of these Steps is nice I think. I didn't debug the actual jsons, so I haven't double checked that.

I added comments about possibly making the usage shorter, but of course those are not critical.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/insights/Opened.java
dmfs: Good idea. Done.

eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/filter/views/EventFilterView.java
dmfs: right, I replaced it with `Toggled`

dmfs:
@lemonboston please check again

lemonboston:
The changes for the `Step`s look good, although `Toggled` could also just delegate to `Selected`, `Unselected` if already available. (Would reduce duplicate definition of path strings "selected", "unselected")

I think the request body logging in the executor decorator could use `Log.d()` instead of `System.out` and moved after the request line log, before the response time log. And probably only logged when content length is > 0. Message could start with the `(4)` like id like the other two, sg like `(4) Request body: ..`.

dmfs:
Ok, I've addressed your comments.

Issue #420, opened 2017-09-15
update UI to show links to the "original" resources, if present


dmfs:
@dmfs commented on [Wed Apr 19 2017](https://github.com/schedjoules/general-issues/issues/182)

In order to get more information, as a user I want to see the original source of some event, like the ones from meetup or facebook.
---

In #181 the API will be updated to also include the "original" source of an event. If present, the UI should show this in a separate line.

lemonboston:
This is a sort of duplicate of: https://github.com/schedjoules/general-issues/issues/170

Issue #254, opened 2017-05-08
handle trivial actions client side


dmfs:
"Add to calendar", "Share" and directions don't really need API support to be implemented. In order to remove them from the API actions result, the SDK should not depend on their presence and generate them locally.

dmfs:
"share" should include a link to the webview

PR #421, opened 2017-09-15
Show sources of events. Implements #420

dmfs:
Also updates to jems 1.8 and removes/replaces some obsolete classes.


eventdiscovery-sdk/src/main/java/com/schedjoules/eventdiscovery/framework/microfragments/eventdetails/fragments/views/EventDetailsView.java
```diff
@@ -76,16 +84,26 @@ public void update(EnrichedEvent enrichedEvent)
...
+                new Mapped<>(
+                        new ActionLinks(actionLinks, ApiLink.Rel.Action.BROWSE),
+                        new ActionFunction(actionFactory, event)),
+                new PresentValues<>(new ArrayIterable<Optional<Action>>(
+                        new OptionalAction(ApiLink.Rel.Action.SHARE, actionLinks, actionFactory, event),
+                        new OptionalAction(ApiLink.Rel.Action.ADD_TO_CALENDAR, actionLinks, actionFactory, event)))))
+        {
+            View actionView = layoutInflater.inflate(R.layout.schedjoules_view_event_details_action, mViews.schedjoulesEventDetailsActionContainer, false);
+            new ActionView(actionView, (TextView) actionView.findViewById(R.id.schedjoules_event_details_action_label)).update(action);
+            mViews.schedjoulesEventDetailsActionContainer.addView(actionView);
```
lemonboston, line 100: Wasn't it the plan to have fix, hardcoded places for each of the actions and not add them dynamically?

Also considering https://github.com/schedjoules/android-event-discovery-sdk/issues/254 wouldn't it be simpler and more future proof to just add this new browse action the same way as it was with the others?

So just adding this to the layout:
```
<include
                            layout="@layout/schedjoules_view_event_details_browse"
                            android:id="@+id/schedjoules_event_details_action_browse"/>
```
And this to the code:
```
        new ActionView(mViews.schedjoulesEventDetailsActionBrowse)
                .update(new OptionalAction(ApiLink.Rel.Action.BROWSE, actionLinks, actionFactory, event));
```
So why the big refactor here creating this loop? Aren't we adding more complexity, and more work for later with this?
The previous state where each action is handled individually seemed more straightforward for me considering the UI plan to have fix places for them, which I suppose could even change, I mean that they could be displayed differently.

dmfs: Note that the place for each action is still hard coded and not determined by the API (which was the initial problem).
Anyway I made this change mostly because there can be multiple browse actions per event. If an event has multiple sources (e.g. Facebook + Meetup), there might be multiple source links. So I needed a loop anyway. Since "Share" and "Add To Calendar" are created the same way, it felt natural to include these in the loop.

lemonboston: I see, I didn't know there can be more than one 'browse' action, I understand the change now.

lemonboston (approved):
Looks good to me, checked the code and in the app as well.

Not important, but isn't the Meetup icon a bit small?
(going to post a screenshot here)

lemonboston:
device-2017-09-18-164458

lemonboston:
(The Meetup icon is bigger in this mockup: https://github.com/schedjoules/general-issues/issues/170#issuecomment-293064509 )

dmfs:
Right, but the icon in the mockup would be larger than 24dp. And the Facebook icon would look way to big at that same size.

Issue #422, opened 2017-09-18
Remove ViewModel library, replace related workaround


lemonboston:
`ViewModel` is only used currently to hold the first page loaded on the splash loading screen in memory, to access it on the list screen. This single usage doesn't justify keeping the library imported, the workaround can be replaced with a simple static variable (which is not a proper solution, but so isn't the current one).

PR #423, opened 2017-09-18
Remove ViewModel and android arch library, use a simple static variab…

lemonboston:
…le as the workaround for the first result page. #422


dmfs:
Can there be any potential race conditions? Would it make sense to synchronize access to the static field?

lemonboston:
I don't think we need to worry about that, even though write and read is from different threads, the read  clearly can only happen later than the write. But I made the field `volatile` to add further guarantee that both threads look at the same memory.
(And in any ways, this is an `Optional` value, there is logic go handle the missing case, that is to load the first page again on the list screen.)

dmfs (approved):

Issue #429, opened 2017-09-25
Change demo app accent color to default


lemonboston:
The accent color in the demo app is set to an orange currently from the original green to test that this color override, if used by partners, works correctly. But this means that our demo and alpha apps has this orange which is not the 'real', default color of the SDK, so we don't see the 'real' version of the app. Even if it's just a color, I suppose that it may affect how we perceive some aspects of the UI.

So we can set the color to the default and only use the override when we are actually testing newly added or modified UI elements.

PR #430, opened 2017-09-25
Disable accent color override in demo app. #429

lemonboston:


dmfs (approved):
I don't consider this an issue, but don't object this change either.

